PR #27, opened 2026-03-11
fix: windows adjustments

AstraBert:
Small adjustments for windows compatibility and setup instructions in the readme


logan-markewich (commented):
This looks good to me -- however, there was some code somewhere that would complain if libreoffic or imagemagik wasn't installed, and it would tell the user how to install it. Might be good to find that text and update it for windows too

PR #30, opened 2026-03-17
add docs support

logan-markewich:
Also cleaned up unused configs


TuanaCelik (commented):
Looks good, just left. couple comments. One thing that I leave to you for now but we seem to often question our layout, I'm wondering if we should go 'overview' (index.md), Agent Skill, reference, and then the rest in 'Guides' or something similar? Up to you.. 

Later we can add an 'examples' section too

PR #42, opened 2026-03-20
support input buffers as escape hatch for URL handling

logan-markewich:
Adds support and tests for reading from buffers/streams in library usage.

This means users can directly stream data through liteparse without writing to disk

**NOTE:** office/image conversion will require disk access. The folder this writes to is controlled by an env var

### DIFF
--- a/src/conversion/convertToPdf.ts
+++ b/src/conversion/convertToPdf.ts
@@ -2,6 +2,7 @@ import { promises as fs, constants as fsConstants } from "fs";
 import { spawn } from "child_process";
 import path from "path";
 import os from "os";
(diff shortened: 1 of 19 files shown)


src/conversion/convertToPdf.ts
```diff
@@ -406,11 +414,59 @@ export async function convertToPdf(
...
     }
   } catch {
     // Ignore cleanup errors
   }
 }
+
+/**
+ * Guess file extension from raw bytes using magic-byte detection.
+ */
+export function guessExtensionFromBuffer(data: Buffer | Uint8Array): string {
```
logan-markewich: too hardcoded, will try to find something else

AstraBert (approved):

PR #105, opened 2026-03-29
fix(conversion): clean up temp dirs for unsupported buffers

### LINKED ISSUE #104: [Bug] convertBufferToPdf leaks temp files for unsupported buffers
## Summary
`convertBufferToPdf()` writes incoming bytes to a temporary directory and then delegates to `convertToPdf()`. If that delegated call returns passthrough content for an unsupported-but-recognized buffer format, the temporary directory is left behind.

## Why this matters
Unsupported buffers that fall back to direct text content still leave temp files on disk even though no converted PDF is produced.

## Reproduction idea
Pass a buffer whose format is recognized by `guessExtensionFromBuffer()` but is not convertible to PDF in `convertToPdf()`, such as a ZIP-like buffer in the current test harness.

## Suggested fix
Clean up the temporary directory in `convertBufferToPdf()` whenever the delegated conversion returns passthrough content or an error instead of a converted PDF path.

Co-Messi:
## Problem
`convertBufferToPdf()` writes input bytes to a temporary directory before delegating to `convertToPdf()`. If the delegated conversion returns passthrough content for an unsupported buffer format, that temporary directory is left behind.

## What changed
- cleaned up the temporary directory in `convertBufferToPdf()` when delegated conversion returns passthrough content or an error
- added a conversion test covering an unsupported-but-recognized buffer format returning passthrough content

## Why this fixes it
This keeps the buffer conversion temp directory tied to the lifetime of the attempted conversion instead of only the successful PDF path.

## Verification
- `npm test -- src/conversion/convertToPdf.test.ts`
- `npm run format:check`
- `npm run build`

Fixes #104

### DIFF
--- a/src/conversion/convertToPdf.test.ts
+++ b/src/conversion/convertToPdf.test.ts
@@ -288,6 +288,22 @@ describe("test convertBufferToPdf", () => {
       force: true,
     });
   });
+
+  it("does not mask passthrough content when temp cleanup fails", async () => {
+    const zipBytes = Buffer.from([0x50, 0x4b, 0x03, 0x04]);
+
+    const { convertBufferToPdf } = await import("./convertToPdf");
+    const fsModule = await import("fs");
+    vi.mocked(fsModule.promises.rm).mockImplementationOnce(async () => {
+      throw new Error("cleanup failed");
+    });
+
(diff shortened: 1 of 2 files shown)


logan-markewich:

Issue #147, opened 2026-04-23
[Feature] Official browser version of LiteParse


simonw:
### Problem Statement

LiteParse works well running directly in a browser with a bit of poking - I have a demo of that here: https://simonw.github.io/liteparse/ - code in this branch: https://github.com/run-llama/liteparse/compare/main...simonw:liteparse:web

### Proposed Solution

I suggest LIteParse adds its own official web build and hosted web version. You're welcome to my vibe-coded starting point if that's useful.

simonw:
More details on my blog: https://simonwillison.net/2026/Apr/23/liteparse-for-the-web/

logan-markewich:
Thank you! Will take a look and see how to best incorporate these changes. Having a web version officially would be fantastic

logan-markewich:
Implement here https://github.com/run-llama/liteparse/pull/151 -- thanks!

PR #148, opened 2026-04-24
Initial prototype of browser-based LiteParse

simonw:
See:
- #147

And https://simonwillison.net/2026/Apr/23/liteparse-for-the-web/

Opening this PR to make it easy to pull in changes if they are useful, but I expect a from-scratch implementation may be a better fit for the project so I won't be at all offended if someone closes this!


logan-markewich:
Really appreciate the work here! I used this as the basis for https://github.com/run-llama/liteparse/pull/151

Its not perfect, but gives users an option. Hopefully can improve in the future!

PR #163, opened 2026-05-09
Rotated page

### LINKED ISSUE #153: [Feature] Rotated page detection
### Problem Statement

PDFs rendered in landscape mode - rotated to 90 or -90 deg. are getting parsed as gibberish

### Proposed Solution

add an optional flag to run tesseract orientation and script detection as a preprocessing step

### Alternatives Considered

_No response_

### Additional Context

_No response_

abhinab-choudhury:
fix #153   added --rotate flag using `pdf-lib`

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "file-type": "^21.3.3",
     "form-data": "^4.0.0",
     "p-limit": "^7.3.0",
+    "pdf-lib": "^1.17.1",
     "sharp": "^0.34.5",
     "tesseract.js": "^7.0.0",
     "unified": "^11.0.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -46,6 +46,7 @@
     "file-type": "^21.3.3",
     "form-data": "^4.0.0",
     "p-limit": "^7.3.0",
+    "pdf-lib": "^1.17.1",
```
logan-markewich, line 49: do we actually need a lib to do this? I would think pdf.js has something for this already?

abhinab-choudhury: Approach 1: CLI --rotate flag
Rotates the PDF at the document level using `pdf-lib` before any processing occurs, both PDF text extraction and OCR operate on the rotated document.

Approach 2: OCR-only rotation change
Would modify the OCR pipeline to rotate images before sending them to the OCR engine.

Is there any other way to solve this!

shivmohith:
We are also facing the same issue. Is there any plan to merge this?

logan-markewich:

PR #165, opened 2026-05-11
feat: misc todos

AstraBert:
- [X] LibreOffice and ImageMagick conversion logic
- [x] HTTP OCR interface


crates/liteparse/src/conversion.rs
```diff
@@ -43,34 +78,387 @@ pub fn is_supported_extension(path: &str) -> bool {
...
+pub fn guess_extension_from_data(data: &[u8]) -> Option<String> {
+    let kind = infer::get(data);
+    let ext = kind.map(|k| k.extension());
+    if let Some(e) = ext {
+        return Some(e.to_string());
+    }
+    None
+}
+
+pub async fn convert_data_to_pdf(
```
logan-markewich, line 453: I actually cant remember if I used async anywhere yet -- we might be able to get away without it? Only debatable place is the http ocr 🤔

AstraBert: Yeah, the main problem was executing with timeouts: the sync version included spawning a thread and maintaining communication between the main and the child thread, so I decided this was easier to read and maintain. But yeah, we can get away with sync and I can refactor this :))

logan-markewich: Ahhh good point on timeouts. If its easier with async lets keep it for now. We can always come back if it ends up being an issue

logan-markewich (approved):

PR #168, opened 2026-05-13
feat: Add OCR-only text mode for corrupted native PDF text layers

shivmohith:
Adds a new `ocrTextMode` parser option for controlling how OCR text is combined with native PDF text on pages where OCR runs.

The default remains:

```json
{
  "ocrTextMode": "merge"
}
```

which preserves the existing behavior.

A new opt-in mode is added:

```json
{
  "ocrTextMode": "ocr-only"
}
```

When enabled, LiteParse uses OCR-derived text items as the source of truth for pages where OCR runs, instead of merging OCR results into the native PDF text layer.

## Motivation

Some PDFs contain a visible page image plus a corrupted or misleading native text layer. In those cases, PDF.js can extract native text items that look structurally valid but produce unreadable output after layout projection.

This creates two problems:

- The corrupted native text remains in `page.textItems`.
- The existing OCR merge path may discard OCR results that overlap native PDF text, assuming the native text is trustworthy.

For these PDFs, OCR successfully reads the visible document, but the final LiteParse output can still be dominated by the bad native text layer.

I cannot attach the source document used to reproduce this because it is proprietary, but the observed behavior was:

```text
Running OCR on pages...
  OCR on page 1...
  Found 0 text items from OCR on page 1
```

The OCR engine had returned results, but they were filtered out because they overlapped existing native PDF text. Using OCR as the only text source for OCR'd pages fixed the extracted output.

## Behavior

`ocrTextMode: "merge"`:

- Existing/default behavior.
- Keeps native PDF text.
- Adds OCR text that passes the existing confidence, garbled-region, and overlap filters.

`ocrTextMode: "ocr-only"`:

- Applies only to pages where OCR actually runs.
- Keeps OCR results above the existing confidence threshold.
- Skips native-text overlap filtering.
- Replaces `page.textItems` with OCR-derived text items for that page.

This does not force OCR to run on every page. It only changes the text-combination behavior after OCR has already been selected for a page.

## CLI

Adds the option to both parse commands:

```bash
lit parse document.pdf --ocr-text-mode ocr-only
lit batch-parse input-dir output-dir --ocr-text-mode ocr-only
```

## Why this is opt-in

The current merge behavior is still useful for normal PDFs where the native text layer is reliable and OCR only fills in missing image text.

`ocr-only` is intended for documents where the native text layer is known to be corrupted, hidden, or less reliable than OCR.

### DIFF
--- a/.changeset/hip-suits-speak.md
+++ b/.changeset/hip-suits-speak.md
@@ -1,0 +1,5 @@
+---
+"@llamaindex/liteparse": minor
+---
+
+Add `ocrTextMode` with an `ocr-only` option for pages where OCR runs. This lets callers ignore native PDF text on OCR'd pages when the document has a corrupted or hidden text layer, while keeping the existing merge behavior as the default.
--- a/README.md
+++ b/README.md
@@ -264,6 +264,7 @@ Options:
   --ocr-server-url <url>  HTTP OCR server URL (uses Tesseract if not provided)
   --no-ocr                Disable OCR
   --ocr-language <lang>   OCR language(s) (default: "en")
+  --ocr-text-mode <mode>  How to combine native text and OCR text (choices: "merge", "ocr-only")
   --num-workers <n>       Number of pages to OCR in parallel (default: CPU cores - 1)
   --max-pages <n>         Max pages to parse (default: "10000")
   --target-pages <pages>  Target pages (e.g., "1-5,10,15-20")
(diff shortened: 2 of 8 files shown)


logan-markewich:
Cool idea! Going to hold off on merging this though. I'm neck deep in a rust port, and would be curious if pdfium has the same issues once that is ready! 

Should be ready in ~1 week

logan-markewich:
#164

logan-markewich:

PR #172, opened 2026-05-15
feat(wasm): add WebAssembly bindings for browser PDF parsing

AstraBert:
Implements browser/WASM support per `crates/WASM_PLAN.md`. Covers steps 1-7 of the plan; steps 8-9 are stubbed out below as follow-ups.

### Core crate prep (steps 1-4)
- **Async `OcrEngine` trait** — now returns `Pin<Box<dyn Future<...>>>`; native impls (`HttpOcrEngine`, `TesseractEngine`) updated; `ocr_merge` awaits results.
- **Async reqwest** — `HttpOcrEngine` uses async `Client` with `.await` instead of blocking.
- **`web-time::Instant`** replaces `std::time::Instant` so timing works on wasm.
- **Feature-gating** — `conversion` and `render` modules are gated `#[cfg(not(target_arch = "wasm32"))]`.
- **tokio features** split: wasm-safe base (`rt`, `macros`, `sync`, `time`, `io-util`); native adds `rt-multi-thread`, `fs`, `process` via target-cfg deps.

### pdfium-sys static link (step 5)
- `crates/pdfium-sys/build.rs` branches on `CARGO_CFG_TARGET_ARCH == "wasm32"`:
  - emits `cargo:rustc-link-lib=static=pdfium`
  - skips the dylib copy and Windows-specific lib handling
  - native path is unchanged

### liteparse-wasm crate (step 6)
- New crate `crates/liteparse-wasm` (cdylib + rlib), added to workspace members.
- `src/lib.rs` (gated to `target_arch = "wasm32"`):
  - `#[wasm_bindgen]` `LiteParse` class with `new(config)`, `parse(Uint8Array) -> Promise<Result>`, `format(result) -> string`, `config` getter
  - camelCase JS config (`ocrLanguage`, `ocrEnabled`, `maxPages`, `targetPages`, `dpi`, `outputFormat`, `preserveVerySmallText`, `password`, `quiet`, `ocrEngine`)
  - JS OCR callback bridge: any object with an async `recognize(imageData, width, height, language)` returning a Promise of `[{text, bbox, confidence}]` is wrapped into a Rust `OcrEngine` impl via `JsFuture`
  - `console_error_panic_hook` wired up behind a default feature
- `OcrEngine` trait: relaxed `Send + Sync` bound on wasm32 (`JsValue` is `!Send`), kept on native.
- Added `LiteParse::with_ocr_engine(Arc<dyn OcrEngine>)` builder; on wasm the parser errors if OCR is enabled without one provided.

### npm package (step 7)
- New `packages/wasm/` published as `@llamaindex/liteparse-wasm@2.0.0`.
- Build scripts wrap `wasm-pack`:
  - `npm run build` → web target
  - `npm run build:bundler` → bundler target (webpack/rollup/vite)
  - `npm run build:nodejs` → node.js target
- README with quick-start, config table, and JS OCR callback example.

## Verification

- ✅ `cargo check -p liteparse` (native)
- ✅ `cargo check -p liteparse-wasm` (native, no-op shell)
- ✅ `cargo check -p liteparse-wasm --target wasm32-unknown-unknown --no-default-features` (with pdfium env stubs)

### Step 8 — End-to-end wasm-pack build
Blocked on upstream: pdfium-binaries does not yet publish a `libpdfium.a` built for `wasm32-unknown-unknown`. `wasm-pack build` will fail at link time until that artifact is available (or we build pdfium ourselves with emscripten). Once available, we wire it in via `PDFIUM_LIB_PATH` / `PDFIUM_INCLUDE_PATH` and verify in a browser.

### Step 9 — CI workflows
Not yet added; intentionally deferred for review. Planned:
- `.github/workflows/ci-wasm.yml` — install wasm-pack, run `wasm-pack build`, upload `pkg/` as an artifact.
- `.github/workflows/release-wasm.yml` — build and `npm publish` `@llamaindex/liteparse-wasm` on tag/dispatch.

## Notes for reviewers

- `liteparse-napi` and `liteparse-python` are untouched and continue to work — the only public API addition on the core crate is `with_ocr_engine(...)`.
- The `parse(path: &str)` method is wasm-gated; WASM users must use `parse_input(PdfInput::Bytes(...))`, which is what the JS `parse(Uint8Array)` binding calls.
- `HttpOcrEngine`'s `multipart` dependency does not compile on wasm32, so the entire `http_simple` module is non-wasm only. The expected wasm OCR story is the JS callback bridge.


logan-markewich:
Added the CICD :)

PR #174, opened 2026-05-16
fix(parser): clean up converted temp PDF when loadDocument fails

### LINKED ISSUE #173: [Bug] Temp conversion directory not cleaned up when loadDocument fails after convert
### Description

When parsing a non-PDF file (e.g. DOCX) or a converted buffer, LiteParse converts the input to a temporary PDF under a `liteparse-*` directory, then calls `pdfEngine.loadDocument()` **before** entering the `try/finally` block that runs `cleanupConversionFiles()`.

If `loadDocument()` throws after conversion succeeds (wrong password, corrupt/invalid PDF from conversion, etc.), the `finally` block never runs and the entire temp directory (converted PDF + LibreOffice/ImageMagick output) is left on disk.

The same pattern exists in `screenshot()`.

**Affected code:** `src/core/parser.ts`, `loadDocument` at lines ~129 and ~154 are outside `try`; cleanup only in `finally` at lines ~211–212 (and ~327–328 for screenshots).

### Steps to Reproduce

1. Ensure LibreOffice (or another supported converter) is installed so a DOCX can be converted to PDF.

2. Use a file that will convert successfully but fail to load, for example:
   - A DOCX that converts to a password-protected PDF, then parse with wrong or missing password, OR
   - A DOCX that converts to a corrupt/invalid PDF that PDF.js cannot open.

3. Run from the CLI (replace the path):

   lit parse C:\path\to\file.docx --password wrong

   Or from the library:

   const parser = new LiteParse();
   await parser.parse("C:\\path\\to\\file.docx", false);

4. The parse should throw (e.g. incorrect password or invalid PDF).

5. Check for leftover temp directories:

   Windows: dir %TEMP%\liteparse-*

   Linux/macOS: ls /tmp/liteparse-*

   (Or check $LITEPARSE_TMPDIR if set.)

6. Confirm a liteparse-* folder still exists and contains the converted .pdf (and any converter output files).

Note: This only happens when conversion succeeds and loadDocument fails. If conversion itself fails, no converted temp PDF is left by this path.

### Error Message

```text
Example when password is wrong (exact text may vary):

Incorrect password for this PDF. Please check the password and try again.

Or when the converted file is not a valid PDF:

Error: Invalid PDF structure.
(depends on PDF.js / load failure)
```

### LiteParse Version

1.5.3

### Operating System

Windows

### Node.js Version

v22.11.0

### Additional Context

**Root cause**

In `parse()`:

1. `convertToPdf()` / `convertBufferToPdf()` creates `mkdtemp(..., 'liteparse-')` and writes the converted PDF there (`src/conversion/convertToPdf.ts`).
2. Parser sets `needsCleanup = true` and `cleanupPath = pdfPath`.
3. `await this.pdfEngine.loadDocument(pdfPath)` runs **outside** `try`.
4. `try/finally` only starts after a successful load; `cleanupConversionFiles(cleanupPath)` is only called in `finally`.

If step 3 throws, step 4 never runs → **no cleanup**.

**When cleanup works**

- Direct `.pdf` path (`needsCleanup === false`): no conversion temp.
- `loadDocument` succeeds; a later step throws: `finally` runs and cleanup works.

**Impact**

- Disk leak per failed load in batch/server workloads.
- No log when cleanup is skipped.

**Suggested fix**

Wrap convert + `loadDocument` in `try/finally` (or outer `finally`) so `cleanupConversionFiles` runs even when load fails; only call `pdfEngine.close(doc)` if `doc` was assigned.

**Related**

Same structure in `screenshot()` (~lines 262–329).

AdemBoukhris457:
Fixes #173 

## Summary

Fixes a resource leak where temporary conversion directories (`liteparse-*`) were left on disk when `loadDocument()` failed after a successful non-PDF → PDF conversion.

`loadDocument()` (and `renderer.loadDocument()` in `screenshot()`) ran **before** the `try/finally` block that calls `cleanupConversionFiles()`. If loading threw (wrong password, corrupt converted PDF, invalid PDF structure, etc.), `finally` never ran and the converted PDF and its temp directory were not removed.

## Changes

- **`src/core/parser.ts`**
  - Resolve the PDF path/buffer first, then run `loadDocument()` inside `try/finally`.
  - Call `pdfEngine.close(doc)` only when `doc` was successfully opened.
  - Apply the same pattern to `screenshot()` (including `PdfiumRenderer.loadDocument()` inside `try`).
- **`src/core/parser.test.ts`**
  - Add tests that assert `cleanupConversionFiles` is called when `loadDocument` rejects after a mocked conversion (parse + screenshot).

## Before / after

| Scenario | Before | After |
|----------|--------|--------|
| Convert DOCX → PDF, then `loadDocument` throws | Temp dir leaked | Temp dir removed in `finally` |
| Convert DOCX → PDF, parse succeeds | Cleaned up | Cleaned up (unchanged) |
| Parse native `.pdf` path directly | No conversion temp | Unchanged |

## Test plan

- [x] `npm test -- src/core/parser.test.ts` (11 tests pass)
- [ ] Manual: convert a DOCX, trigger a load failure (e.g. wrong `--password`), confirm no leftover `liteparse-*` dirs under `$TMP` / `%TEMP%`

### DIFF
--- a/src/core/parser.test.ts
+++ b/src/core/parser.test.ts
@@ -345,6 +345,8 @@ const {
 
 import { LiteParse } from "./parser";
 import { LiteParseConfig, ScreenshotResult } from "./types";
+import { cleanupConversionFiles } from "../conversion/convertToPdf.js";
+import { PdfJsEngine } from "../engines/pdf/pdfjs.js";
 
 vi.mock("../conversion/convertToPdf.js", async () => {
   const actual = await vi.importActual<typeof import("../conversion/convertToPdf.js")>(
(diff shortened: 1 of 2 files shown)


logan-markewich:

PR #180, opened 2026-05-16
fix(parser): serialize parse and screenshot on shared LiteParse instance

### LINKED ISSUE #179: [Bug] LiteParse is not safe for concurrent parse() or screenshot() on one instance
### Description

A single `LiteParse` instance is not safe for overlapping `parse()` or `screenshot()` calls. The class shares one `PdfJsEngine` (and one OCR engine) with **mutable instance state** (`currentPdfPath`, `currentPdfData`, shared `PdfiumRenderer`). Concurrent calls can read or close the wrong document.

There is no mutex, queue, or per-call engine isolation, and this limitation is not documented in the public API.

**Affected code:**
- `src/core/parser.ts` — one `pdfEngine` / `ocrEngine` per `LiteParse`; `terminate()` on shared Tesseract in `finally`
- `src/engines/pdf/pdfjs.ts` — `currentPdfPath`, `currentPdfData`, `pdfiumRenderer` shared across calls

### Steps to Reproduce

**1.** Use LiteParse with OCR enabled (optional, but makes races easier to see).

**2.** Create **one** `LiteParse` instance and parse two PDFs **in parallel**:

```javascript

const parser = new LiteParse({ ocrEnabled: true });

const [resultA, resultB] = await Promise.all([
  parser.parse("/path/to/document-a.pdf"),
  parser.parse("/path/to/document-b.pdf"),
]);
```

**3.** Compare with **sequential** parsing (expected correct behavior):

```javascript
const parserA = new LiteParse({ ocrEnabled: true });
const parserB = new LiteParse({ ocrEnabled: true });

const sequentialA = await parserA.parse("/path/to/document-a.pdf");
const sequentialB = await parserB.parse("/path/to/document-b.pdf");
```

**4.** Run step 2 several times. Parallel results may differ from step 3 or change between runs.

---

**Alternative (server-style repro):**

**1.** Create a single shared parser:

```javascript
const parser = new LiteParse({ ocrEnabled: true });
```

**2.** Simulate two concurrent requests:

```javascript
await Promise.all([
  parser.parse("/path/to/file-a.pdf"),
  parser.parse("/path/to/file-b.pdf"),
]);
```

**3.** Observe wrong/mixed text or intermittent errors.

---

**Note:** Sequential use on the same instance is usually fine:

```javascript
await parser.parse("/path/to/document-a.pdf");
await parser.parse("/path/to/document-b.pdf");
```

The bug appears when both calls **overlap in time**, not when they run one after another.

### Error Message

```text
Often **no error** — calls may succeed with incorrect or mixed text.

Sometimes intermittent failures when one call's `finally` runs while another is still active, for example:
- OCR / render errors
- Incomplete text
- Exceptions from closed PDFium renderer or terminated Tesseract workers
```

### LiteParse Version

1.5.3

### Operating System

Windows

### Node.js Version

v22.11.0

### Additional Context

**Why it happens**

| Component | Shared state / behavior |
|-----------|------------------------|
| `PdfJsEngine` | `currentPdfPath`, `currentPdfData`, `pdfiumRenderer` overwritten per `loadDocument()` |
| OCR rendering | `renderPageImage()` uses whichever PDF is currently loaded |
| `parse()` `finally` | `pdfEngine.close(doc)` and `tesseract.terminate()` affect the shared engines |

**Example timeline**

1. `parse(docA)` loads A into the engine.
2. `parse(docB)` loads B → engine now points at B.
3. `parse(docA)` still running OCR → may render pages from B.
4. `parse(docB)` finishes → `close()` / `terminate()` may break A mid-flight.

**When it does NOT happen**

- `await parse(a); await parse(b);` on the same instance (fully sequential)
- `new LiteParse()` per request / per job

**When it DOES happen**

- `Promise.all([parser.parse(a), parser.parse(b)])`
- HTTP server or worker pool with one global `LiteParse` instance
- Overlapping `parse()` and `screenshot()` on the same instance

**Impact**

- Wrong document text in production APIs (silent data corruption)
- Flaky OCR / parse failures under load
- Hard to reproduce because it depends on timing

**Suggested fix directions**

1. Document: one in-flight operation per `LiteParse` instance; create a new instance per concurrent job.
2. Add an internal mutex/queue so concurrent calls are serialized on one instance.
3. Use per-call `PdfJsEngine` (and isolate OCR lifecycle) instead of shared instance state.
4. Avoid calling `tesseract.terminate()` after every parse if the instance is reused.

**Related**

Independent of conversion/temp-file bugs (#1–#3). This is orchestration / shared engine state in Section 1 (`parser.ts` + `pdfjs.ts`).

AdemBoukhris457:
Fixes #179 

## Summary

Fixes unsafe concurrent use of a single `LiteParse` instance. Overlapping `parse()` or `screenshot()` calls shared one `PdfJsEngine` (and optional OCR engine) with mutable state, which could load the wrong PDF, return mixed text, or fail when one call closed resources another still needed.

Operations on the same instance are now **serialized** with an internal lock.

## Problem

`LiteParse` reused:

- `PdfJsEngine` — `currentPdfPath`, `currentPdfData`, shared `PdfiumRenderer`
- `TesseractEngine` (optional) — terminated in `parse()` `finally` while another call could still run

Example failure mode:

```javascript
await Promise.all([
  parser.parse("document-a.pdf"),
  parser.parse("document-b.pdf"),
]);
```

Could succeed with **wrong text** or **intermittent errors** depending on timing.

## Changes

- **`src/core/parser.ts`**
  - Add `runExclusive()` queue backed by `operationLock`
  - Public `parse()` / `screenshot()` run inside the lock
  - Implementation moved to private `parseDocument()` / `screenshotDocument()`
  - Document that concurrent calls on one instance are queued
- **`src/core/parser.test.ts`**
  - Test that overlapping `parse()` calls never have more than one active `loadDocument` at a time

## Behavior

| Before | After |
|--------|--------|
| Concurrent calls on one instance → races | Calls run one at a time (FIFO queue) |
| `new LiteParse()` per request | Still recommended for parallel throughput |
| Sequential `await parse(a); await parse(b)` | Unchanged (still works) |

**Note:** This does **not** make one instance faster in parallel — it makes it **correct**. For real parallelism, use separate `LiteParse` instances per job.

### DIFF
--- a/src/core/parser.test.ts
+++ b/src/core/parser.test.ts
@@ -345,6 +345,8 @@ const {
 
 import { LiteParse } from "./parser";
 import { LiteParseConfig, ScreenshotResult } from "./types";
+import { PdfJsEngine } from "../engines/pdf/pdfjs.js";
+import { projectPagesToGrid } from "../processing/grid.js";
 
 vi.mock("../conversion/convertToPdf.js", async () => {
   const actual = await vi.importActual<typeof import("../conversion/convertToPdf.js")>(
@@ -616,5 +618,35 @@ describe("test screenshot", () => {
       mockScreenshotResults[2],
       mockScreenshotResults[3],
     ]);
+  });
+});
+
+describe("concurrent operations", () => {
+  it("serializes overlapping parse() calls on one instance", async () => {
+    let active = 0;
+    let maxActive = 0;
+
+    vi.mocked(projectPagesToGrid).mockReset();
+    vi.mocked(projectPagesToGrid).mockResolvedValue(structuredClone(mockParsedPages));
+
+    vi.mocked(PdfJsEngine).mockImplementationOnce(
+      class {
+        loadDocument = vi.fn(async () => {
+          active += 1;
+          maxActive = Math.max(maxActive, active);
+          await new Promise((resolve) => setTimeout(resolve, 30));
+          active -= 1;
+          return mockPdfDocument;
+        });
+        extractAllPages = vi.fn().mockResolvedValue(mockPages);
+        extractPage = vi.fn().mockResolvedValue(mockPages[0]);
+        renderPageImage = vi.fn();
+        close = vi.fn();
+      } as unknown as typeof PdfJsEngine
+    );
+
+    const parser = new LiteParse({ ocrEnabled: false, outputFormat: "text" });
+    await Promise.all([parser.parse("/tmp/a.pdf"), parser.parse("/tmp/b.pdf")]);
+    expect(maxActive).toBe(1);
   });
 });
--- a/src/core/parser.ts
+++ b/src/core/parser.ts
@@ -59,6 +59,8 @@ export class LiteParse {
   private config: LiteParseConfig;
   private pdfEngine: PdfEngine;
   private ocrEngine?: OcrEngine;
+  /** Ensures only one parse/screenshot uses shared engines at a time */
+  private operationLock: Promise<void> = Promise.resolve();
 
   /**
    * Create a new LiteParse instance.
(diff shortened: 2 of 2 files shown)


logan-markewich:

PR #183, opened 2026-05-16
refactor(parser): build result.json with buildJSON instead of stringify/parse

AdemBoukhris457:
## Summary

Removes an unnecessary JSON round-trip when `outputFormat` is `"json"`.

Previously, `parse()` built `result.json` via:

```ts
result.json = JSON.parse(formatJSON(result));
```

`formatJSON` already calls `buildJSON` and then `JSON.stringify` for pretty printing. Parsing that string back into an object added extra CPU and memory on every JSON parse, with no benefit for library callers.

Now:

```ts
result.json = buildJSON(result.pages);
```

`formatJSON` is unchanged and remains the right API for producing a formatted JSON **string** (e.g. CLI file output still uses `JSON.stringify(result.json, null, 2)`).

## Why

- Avoid redundant serialize → parse on large documents
- Clearer separation: `buildJSON` for objects, `formatJSON` for strings
- Same output shape for normal parse data (existing tests unchanged)

### DIFF
--- a/src/core/parser.ts
+++ b/src/core/parser.ts
@@ -15,7 +15,7 @@ import { TesseractEngine } from "../engines/ocr/tesseract.js";
 import { HttpOcrEngine } from "../engines/ocr/http-simple.js";
 import { projectPagesToGrid } from "../processing/grid.js";
 import { buildBoundingBoxes } from "../processing/bbox.js";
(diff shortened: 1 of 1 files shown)


logan-markewich:

PR #189, opened 2026-05-17
fix: reuse Tesseract worker pool across parse() calls

### LINKED ISSUE #188: [Bug] Tesseract OCR workers are terminated after every parse()
### Description

`LiteParse` is designed as a long-lived instance: the constructor creates a `TesseractEngine` when `ocrEnabled` is true and no `ocrServerUrl` is set. Callers commonly reuse one parser for many documents (batch scripts, servers, notebooks).

However, **`parse()` always calls `TesseractEngine.terminate()` in its `finally` block**, destroying the worker pool and scheduler after every run—success or failure, whether or not OCR actually ran on that document.

### Current behavior

1. **Constructor** (`src/core/parser.ts`) — Creates `TesseractEngine` but does not start workers yet (lazy init).

2. **First `parse()` that needs OCR** — `TesseractEngine.recognize()` → `initialize(language)` spawns `numWorkers` workers (default 4), loads language data (may download from CDN if no local `tessdata`).

3. **`parse()` `finally` block** — Always runs:

   ```typescript
   if (this.ocrEngine && "terminate" in this.ocrEngine) {
     await (this.ocrEngine as TesseractEngine).terminate();
   }
   ```

4. **Next `parse()` on the same instance** — Workers are gone; `initialize()` runs again from scratch (new scheduler, new workers, reload language).

`TesseractEngine.terminate()` (`src/engines/ocr/tesseract.ts`) clears the scheduler and worker array:

```typescript
async terminate(): Promise<void> {
  if (this.scheduler) {
    await this.scheduler.terminate();
    this.scheduler = undefined;
  }
  this.workers = [];
  this.currentLanguage = undefined;
}
```

### What is *not* affected

| Configuration | Behavior |
|---------------|----------|
| `ocrEnabled: false` | No `ocrEngine`; no `terminate()` |
| `ocrServerUrl` set (HTTP OCR) | `HttpOcrEngine` has no `terminate()`; pool not torn down |
| `screenshot()` only | Does not call `terminate()` (asymmetric with `parse()`) |
| One `parse()` then process exit | Overhead is less visible |

### Impact

| Area | Effect |
|------|--------|
| **Batch processing** | `for (const f of files) await parser.parse(f)` pays full Tesseract startup per file (often seconds each on cold start) |
| **Servers / singletons** | One shared `LiteParse` instance cannot keep a warm OCR pool |
| **Network / offline** | Without `TESSDATA_PREFIX` / `tessdataPath`, repeated init may re-hit CDN downloads |
| **UX vs expectation** | API shape (`new LiteParse()` once, many `parse()` calls) implies reuse; implementation defeats it for built-in OCR |
| **Resource churn** | Spawn/teardown of multiple workers every document |

This is **not** a text-extraction correctness bug—results are typically correct. It is a **lifecycle / performance** bug relative to how the library is meant to be used.

### Expected behavior

- Keep Tesseract workers alive across `parse()` calls on the same `LiteParse` instance.
- Tear down workers only when the user explicitly disposes the parser (e.g. `await parser.destroy()`), or when the process exits (CLI one-shot is fine).
- `initialize()` already short-circuits if the scheduler exists and language is unchanged; that optimization is nullified by `terminate()` after every `parse()`.

### Suggested fix direction

1. Remove `terminate()` from `parse()`'s `finally` block in `src/core/parser.ts`.
2. Add `LiteParse.destroy()` (or `close()`) that calls `pdfEngine` cleanup and `ocrEngine.terminate()` when the user is done.
3. Document lifecycle in README / `src/core/README.md` (currently notes "terminates OCR workers" after parsing—which encodes the problematic behavior).
4. Optional: call `destroy()` from CLI after a single run if needed for clean process exit.

### Affected code

| Location | Behavior |
|----------|----------|
| `src/core/parser.ts` — `parse()` `finally` | Calls `TesseractEngine.terminate()` every time |
| `src/engines/ocr/tesseract.ts` — `initialize` / `terminate` | Pool built on demand; `terminate()` wipes state |
| `src/engines/ocr/README.md` | Documents `terminate()` as "called by LiteParse after parsing" |

### Steps to Reproduce

1. Install LiteParse with built-in OCR (default: `ocrEnabled: true`, no `ocrServerUrl`).
2. Use a PDF that triggers OCR (scanned page, sparse text, or embedded image depending on config).
3. Reuse one parser for two parses and compare wall time:

   ```typescript

   const parser = new LiteParse({ ocrEnabled: true, numWorkers: 4 });

   console.time("parse-1");
   await parser.parse("path/to/ocr-heavy.pdf");
   console.timeEnd("parse-1");

   console.time("parse-2");
   await parser.parse("path/to/another-ocr-heavy.pdf");
   console.timeEnd("parse-2");
   // Second parse still pays full worker init (~seconds), not incremental OCR only
   ```

4. Observe in logs/debugging (or by timing): each `parse()` ends with worker teardown; the second `parse()` recreates the full worker pool before OCR runs.

**Minimal repro without OCR-heavy PDF:** Enable OCR and call `parse()` twice on any PDF—even if OCR does not run, `terminate()` still executes when `ocrEngine` is Tesseract (workers may never have been created, but the pattern still forces re-init on the next OCR need).

### Error Message

```text

```

### LiteParse Version

1.5.3

### Operating System

Windows

### Node.js Version

v22.11.0

### Additional Context

- `TesseractEngine.initialize()` already avoids redundant work when `scheduler` exists and language matches—but `parse()` always calls `terminate()` first via `finally`, so this fast path rarely helps across documents.
- `terminate()` inside `initialize()` when **language changes** is appropriate; per-parse teardown is not.
- Related audit item: reusing `numWorkers` for both Tesseract pool size and page-level concurrency (`pLimit`) is a separate issue (#10).
- HTTP OCR users are unaffected.

AdemBoukhris457:
Fixes #188 

## Summary

- Stop calling `TesseractEngine.terminate()` in `parse()`'s `finally` block so a single `LiteParse` instance keeps a warm OCR worker pool across documents.
- Add `LiteParse.destroy()` for explicit shutdown when the parser is no longer needed.
- Call `parser.destroy()` from CLI `parse`, `screenshot`, and `batch` commands so one-shot runs still release workers on exit.

Fixes the performance issue where every `parse()` tore down and recreated all Tesseract workers (often adding seconds per file in batch jobs). HTTP OCR is unchanged (`HttpOcrEngine` has no `terminate()`).

## Motivation

`LiteParse` is constructed once and reused for many files (batch scripts, servers, notebooks). The previous lifecycle called `terminate()` after every `parse()`, defeating `TesseractEngine.initialize()`'s worker caching and forcing full pool recreation on each document.

## API change

**New:** `await parser.destroy()` — terminates the built-in Tesseract worker pool when done.

**Behavior change:** `parse()` no longer terminates OCR workers automatically. Callers that need cleanup should call `destroy()` (CLI does this automatically).

### DIFF
--- a/cli/parse.ts
+++ b/cli/parse.ts
@@ -172,47 +172,47 @@ program
       const parser = new LiteParse(config);
 
       try {
-      // Read from stdin or file
-      let input: string | Buffer;
-      if (isStdin) {
-        const chunks: Buffer[] = [];
-        for await (const chunk of process.stdin) {
-          chunks.push(chunk);
+        // Read from stdin or file
+        let input: string | Buffer;
+        if (isStdin) {
+          const chunks: Buffer[] = [];
+          for await (const chunk of process.stdin) {
+            chunks.push(chunk);
+          }
+          input = Buffer.concat(chunks);
+          if (input.length === 0) {
+            console.error("Error: No data received from stdin");
+            process.exit(1);
+          }
+        } else {
+          input = file;
         }
-        input = Buffer.concat(chunks);
-        if (input.length === 0) {
-          console.error("Error: No data received from stdin");
-          process.exit(1);
+
+        // Parse document (quiet flag controls progress output)
+        const result = await parser.parse(input, quiet);
+
+        // Format output based on format
+        let output: string;
+        switch (config.outputFormat) {
+          case "json":
+            output = JSON.stringify(result.json, null, 2);
+            break;
+          case "text":
+          default:
+            output = result.text;
+            break;
         }
-      } else {
-        input = file;
-      }
 
-      // Parse document (quiet flag controls progress output)
-      const result = await parser.parse(input, quiet);
-
-      // Format output based on format
-      let output: string;
-      switch (config.outputFormat) {
-        case "json":
-          output = JSON.stringify(result.json, null, 2);
-          break;
-        case "text":
-        default:
-          output = result.text;
-          break;
-      }
-
-      // Write to file or stdout
-      if (options.output) {
-        await fs.writeFile(options.output, output);
-        if (!quiet) {
-          console.error(`\n✓ Parsed ${result.pages.length} pages → ${options.output}`);
+        // Write to file or stdout
+        if (options.output) {
+          await fs.writeFile(options.output, output);
+          if (!quiet) {
+            console.error(`\n✓ Parsed ${result.pages.length} pages → ${options.output}`);
+          }
+        } else {
+          // Output result to stdout (can be piped)
+          console.log(output);
         }
-      } else {
-        // Output result to stdout (can be piped)
-        console.log(output);
-      }
       } finally {
         await parser.destroy();
       }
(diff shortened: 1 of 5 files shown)


logan-markewich:

PR #204, opened 2026-05-21
chore: vbump all packages

AstraBert:


packages/node/package.json
logan-markewich, line 1: I thiiiink youll also need to bump the optional deps in here two, the one for each binary

PR #208, opened 2026-05-22
fix: scan output_dir for PDF instead of assuming LibreOffice preserves filename stem

### LINKED ISSUE #207: [Bug] DOCX conversion fails for filenames with special characters (parentheses, leading digits)
## Summary

`convert_office_document()` in `crates/liteparse/src/conversion.rs` fails with **"LibreOffice conversion succeeded but output PDF not found"** when the input file has a name that LibreOffice sanitises during conversion.

## Root cause

After running `soffice --convert-to pdf --outdir <tmpdir> <file>`, the code constructs the expected output path as:

```rust
let pdf_path = Path::new(output_dir).join(format!("{base_name}.pdf"));
```

where `base_name` is the stem of the **original** filename. However, LibreOffice sanitises filenames before writing the output PDF — it strips parentheses, leading digit sequences, spaces, and other special characters. The file it produces therefore has a **different stem**, so the constructed path never exists.

**Example:**
- Input: `52304751_AnuragLahare_E2 (1).docx`
- Expected by liteparse: `52304751_AnuragLahare_E2 (1).pdf`
- Actual output by LibreOffice: `AnuragLahare_E2_1_.pdf`

## Reproducer

```bash
# Any DOCX whose name contains parentheses or starts with digits triggers this
lit parse "52304751_AnuragLahare_E2 (1).docx"
# → Error: LibreOffice conversion succeeded but output PDF not found
```

## Fix

Since `output_dir` is always a **fresh** `tempfile::TempDir` that contains exactly one file after a successful conversion, scan it for any `.pdf` entry instead of assuming the filename was preserved:

```rust
let mut read_dir = tokio::fs::read_dir(output_dir).await
    .map_err(|e| LiteParseError::Conversion(format!("cannot read output dir: {e}")))?;
while let Some(entry) = read_dir.next_entry().await
    .map_err(|e| LiteParseError::Conversion(format!("error reading output dir: {e}")))?
{
    let path = entry.path();
    if path.extension().map(|e| e.eq_ignore_ascii_case("pdf")).unwrap_or(false) {
        return Ok(path.to_string_lossy().to_string());
    }
}
Err(LiteParseError::Conversion(
    "LibreOffice conversion succeeded but output PDF not found".into(),
))
```

A PR with this fix is attached.

kiran-kumar95:
Fixes #207

## Problem

`convert_office_document()` fails with *"LibreOffice conversion succeeded but output PDF not found"* for any DOCX whose filename contains characters that LibreOffice sanitises during conversion (parentheses, leading digit sequences, spaces, etc.).

The old code constructs the expected output path from the **original** filename stem:

```rust
let pdf_path = Path::new(output_dir).join(format!("{base_name}.pdf"));
```

But LibreOffice rewrites the filename — e.g. `52304751_AnuragLahare_E2 (1).docx` becomes `AnuragLahare_E2_1_.pdf` — so the path check always fails.

## Fix

Since `output_dir` is a **fresh** `tempfile::TempDir` per conversion call and will contain exactly one file on success, the fix simply scans `output_dir` for any `.pdf` entry and returns its path. No filename assumptions are made.

## Testing

Verified locally with:
- `52304751_AnuragLahare_E2 (1).docx` (leading digits + parentheses)
- `Vaibhav_Dixit_Latest (1).docx` (parentheses only)

Both previously failed; both now parse successfully via `lit parse`.

### DIFF
--- a/crates/liteparse/src/conversion.rs
+++ b/crates/liteparse/src/conversion.rs
@@ -382,22 +382,31 @@ pub async fn convert_office_document(
 
     execute_command(&libre_office_cmd, args, 120_000).await?;
 
-    let base_name = Path::new(file_path)
-        .file_stem()
-        .and_then(|s| s.to_str())
-        .ok_or_else(|| LiteParseError::Conversion(format!("invalid file path: {file_path}")))?;
-    let pdf_path = Path::new(output_dir)
-        .join(format!("{base_name}.pdf"))
-        .to_string_lossy()
-        .to_string();
-
-    if tokio::fs::metadata(&pdf_path).await.is_ok() {
-        Ok(pdf_path)
-    } else {
-        Err(LiteParseError::Conversion(
-            "LibreOffice conversion succeeded but output PDF not found".into(),
-        ))
+    // LibreOffice may sanitise the input filename (e.g. strip parentheses, leading
+    // digits, spaces) so the expected `<file_stem>.pdf` may not match what was
+    // actually produced.  Since `output_dir` is always a fresh temp directory that
+    // contains exactly one file after a successful conversion, scanning it for any
+    // `.pdf` entry is the most robust approach.
+    let mut read_dir = tokio::fs::read_dir(output_dir)
+        .await
+        .map_err(|e| LiteParseError::Conversion(format!("cannot read output dir: {e}")))?;
+    while let Some(entry) = read_dir
+        .next_entry()
+        .await
+        .map_err(|e| LiteParseError::Conversion(format!("error reading output dir: {e}")))?
+    {
+        let path = entry.path();
+        if path
+            .extension()
+            .map(|e| e.eq_ignore_ascii_case("pdf"))
+            .unwrap_or(false)
+        {
+            return Ok(path.to_string_lossy().to_string());
+        }
     }
+    Err(LiteParseError::Conversion(
+        "LibreOffice conversion succeeded but output PDF not found".into(),
+    ))
 }
 
 /// Convert images to PDF using ImageMagick.
--- a/crates/liteparse/src/conversion.rs
+++ b/crates/liteparse/src/conversion.rs
@@ -381,12 +381,17 @@ pub async fn convert_office_document(
     args.push(file_path);
 
     execute_command(&libre_office_cmd, args, 120_000).await?;
+    find_pdf_in_dir(output_dir).await
+}
 
-    // LibreOffice may sanitise the input filename (e.g. strip parentheses, leading
-    // digits, spaces) so the expected `<file_stem>.pdf` may not match what was
-    // actually produced.  Since `output_dir` is always a fresh temp directory that
-    // contains exactly one file after a successful conversion, scanning it for any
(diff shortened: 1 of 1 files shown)


kiran-kumar95:
Hi! This is a one-function fix for a bug where LibreOffice filename sanitisation causes conversion to fail silently for files with special characters. Happy to add a test case if that would help get this reviewed. Could a maintainer approve the workflows to run CI? Thanks!

To give a fuller picture of the scope — here are examples of filenames that all trigger the same failure, because LibreOffice sanitises them before writing the output PDF:

| Input filename | Expected by liteparse | Actual LibreOffice output |
|---|---|---|
| `52304751_AnuragLahare_E2 (1).docx` | `52304751_AnuragLahare_E2 (1).pdf` | `AnuragLahare_E2_1_.pdf` |
| `Vaibhav_Dixit_Latest (1).docx` | `Vaibhav_Dixit_Latest (1).pdf` | `Vaibhav_Dixit_Latest_1_.pdf` |
| `Report (Final Draft).docx` | `Report (Final Draft).pdf` | `Report_Final_Draft_.pdf` |
| `001_Onboarding Checklist.docx` | `001_Onboarding Checklist.pdf` | `Onboarding_Checklist.pdf` |
| `Q1 Sales & Revenue.docx` | `Q1 Sales & Revenue.pdf` | `Q1_Sales___Revenue.pdf` |

Any filename with parentheses, leading digit sequences, spaces, ampersands, or other non-alphanumeric characters hits this. The fix (scan the temp dir for any `.pdf` file) handles all of them in one go since the temp dir always contains exactly one file after a successful conversion.

kiran-kumar95:
**Update: contribution checklist now fully satisfied** ✅

Following the CONTRIBUTING.md guidelines:

| Check | Result |
|---|---|
| `cargo fmt --check` | ✅ passes (FMT OK) |
| `cargo clippy` | ✅ passes — 0 new warnings from our changes (2 pre-existing warnings in `ocr/mod.rs` unrelated to this PR) |
| Tests added | ✅ 2 unit tests — run without LibreOffice, finish in <1 ms |
| Single focused change | ✅ one function fixed, one helper extracted |

**What the second commit adds:**

The scanning logic is extracted into a private `find_pdf_in_dir()` helper so it can be unit-tested independently of LibreOffice.  Two tests cover the full behaviour:

```
test conversion::tests::test_find_pdf_in_dir_returns_sanitised_name ... ok
test conversion::tests::test_find_pdf_in_dir_empty_dir_returns_error ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; finished in 0.00s
```

The first test directly reproduces the bug: it places a PDF with the LibreOffice-sanitised name (`AnuragLahare_E2_1_.pdf`) in a temp dir — the name LibreOffice actually produced from `52304751_AnuragLahare_E2 (1).docx` — and asserts it is found. The old code would have looked for `52304751_AnuragLahare_E2 (1).pdf` and missed it entirely.

logan-markewich (approved):
This makes sense to me, thanks!

PR #211, opened 2026-05-23
fix: clean up buffer conversion temp dirs on failure paths

### LINKED ISSUE #210: [Bug] Buffer parse leaves staging temp directories on disk when conversion or parsing fails
### Description

When `LiteParse.parse()` receives **raw bytes** (`Buffer` / `Uint8Array` in Node, `parse_bytes()` in Python, or `PdfInput::Bytes` in Rust), the conversion layer writes the upload to a temporary staging directory, optionally converts it to PDF, then parses. Cleanup of that staging directory is **manual and incomplete** — it only runs on specific success paths. When conversion or parsing fails, orphaned `liteparse-*` directories remain on disk, each containing a full copy of the uploaded file (e.g. `input.docx`, `input.pdf`, `input.bin`).

This affects server/API workloads that parse in-memory uploads. It is **not** a parsing correctness bug; parsed output is usually fine on success. The issue is **temp file hygiene** (disk usage, privacy, silent accumulation under load).

### Root cause

In `crates/liteparse/src/conversion.rs`, `convert_data_to_pdf` creates a staging directory and calls `TempDir::keep()`, which disables automatic RAII cleanup:

```rust
let tmp_dir = tempfile::Builder::new().prefix("liteparse-").tempdir()?;
let tmp_path = tmp_dir
    .keep()
    .join(format!("input.{}", ext.unwrap_or("bin".to_string())));
tokio::fs::write(&tmp_path, data).await?;
convert_to_pdf(tmp_path.to_str().unwrap(), password, true).await
```

After `keep()`, the staging directory must be deleted manually elsewhere. If that manual cleanup does not run, the directory leaks.

### Two temp directories per buffer conversion

| Directory | Created by | Contains | Example |
|-----------|------------|----------|---------|
| **A — staging** | `convert_data_to_pdf` | Original upload | `/tmp/liteparse-abc123/input.docx` |
| **B — output** | `convert_to_pdf` | Converted PDF | `/tmp/liteparse-def456/document.pdf` |

### When cleanup runs today

| Scenario | Dir A (staging) | Dir B (output) |
|----------|-----------------|----------------|
| Non-PDF buffer, conversion **succeeds**, parse **succeeds** | Removed in `convert_to_pdf` (line ~130) | Removed in `parser.rs` (line ~206) |
| Non-PDF buffer, conversion **fails** | **Leaks** | Usually cleaned (TempDir RAII on error) |
| PDF buffer, parse **succeeds** | Removed in `parser.rs` (line ~206) | N/A |
| PDF buffer, parse **fails** (extract/OCR error) | **Leaks** | N/A |
| Unknown bytes → `input.bin` → unsupported format | **Leaks** | N/A |

PDF bytes take an early return in `convert_to_pdf` that **skips** the staging cleanup block:

```rust
if ext == "pdf" {
    return Ok((ConversionResult { pdf_path: path.to_string(), ... }, None));
}
```

Parser cleanup at the end of `parse_input` only runs on the **success path** — there is no `try/finally` or `Drop` guard:

```rust
if is_converted && let PdfInput::Path(p) = validated_input {
    std::fs::remove_dir_all(Path::new(&p).parent().unwrap())?;
}
Ok(ParseResult { ... })
```

Any error before that line (failed extract, OCR, invalid pages, etc.) exits early and never cleans the staging directory.

### When it happens

| Input | Affected? |
|-------|-----------|
| `parse(Buffer)` / `parse(Uint8Array)` for non-PDF (DOCX, PNG, …) on **conversion failure** | **Yes** |
| `parse(Buffer)` for PDF bytes on **parse failure** | **Yes** |
| `parse_bytes()` in Python on same paths | **Yes** |
| `parse("file.docx")` file path | **No** — no buffer staging dir |
| `parse(pdfBuffer)` on **success** | **No** — cleaned on success |

### Impact

| Area | Effect |
|------|--------|
| **Disk** | Full duplicate of every failed buffer upload left in temp |
| **Privacy** | Sensitive documents linger until OS temp cleanup |
| **Servers** | High traffic → many orphaned `liteparse-*` dirs; disk pressure |
| **Silent** | No error is thrown; easy to miss in development |

### Expected behavior

After `parse()` / `parse_bytes()` completes — **success or failure** — all LiteParse-created temp directories for that request should be removed.

Possible fixes:
- Use a single scoped parent `TempDir` for input + output and rely on RAII (`Drop`)
- Add `try/finally`-style cleanup in `parse_input` and `convert_data_to_pdf`
- Avoid `keep()` unless paired with guaranteed cleanup on all exit paths
- For PDF bytes, detect `%PDF` magic and skip disk staging entirely (`PdfInput::Bytes` fast path)

### Affected code

| Location | Behavior |
|----------|----------|
| `crates/liteparse/src/conversion.rs` — `convert_data_to_pdf` | Creates staging dir A with `keep()`; no guaranteed cleanup |
| `crates/liteparse/src/conversion.rs` — `convert_to_pdf` | Cleans dir A only when `is_temporary_path &&` conversion succeeds; PDF early return skips cleanup |
| `crates/liteparse/src/parser.rs` — `PdfInput::Bytes` branch | Cleans staging dir only on successful parse (line ~206) |

### Repro A — conversion failure (non-PDF buffer)

1. Install LiteParse (`@llamaindex/liteparse` or `liteparse` Python package).

2. **Do not** install ImageMagick (or use corrupt/invalid image bytes).

3. Parse a PNG buffer and expect conversion to fail:

   ```typescript
   import { readFile, readdir } from "fs/promises";
   import os from "os";

   const before = (await readdir(os.tmpdir())).filter((e) => e.startsWith("liteparse-"));

   const png = await readFile("receipt.png"); // or any PNG bytes
   const parser = new LiteParse({ ocrEnabled: false });

   try {
     await parser.parse(png);
   } catch {
     // expected if ImageMagick unavailable
   }

   const after = (await readdir(os.tmpdir())).filter((e) => e.startsWith("liteparse-"));
   console.log("new liteparse temp dirs:", after.length - before.length);
   // Inspect new dirs — each contains input.png
   ```

4. Repeat several times; orphaned directories accumulate.

### Repro B — parse failure (PDF buffer)

1. Install LiteParse with a working PDF parse setup.

2. Parse invalid/truncated PDF bytes (or PDF that fails extract/OCR):

   ```typescript
   import { readdir } from "fs/promises";
   import os from "os";

   const badPdf = Buffer.from("%PDF-1.4 truncated");
   const parser = new LiteParse({ ocrEnabled: true });

   const before = (await readdir(os.tmpdir())).filter((e) => e.startsWith("liteparse-"));

   try {
     await parser.parse(badPdf);
   } catch {
     // expected parse failure
   }

   const after = (await readdir(os.tmpdir())).filter((e) => e.startsWith("liteparse-"));
   console.log("new liteparse temp dirs:", after.length - before.length);
   // Each orphan dir contains input.pdf with the uploaded bytes
   ```

3. Or use a valid PDF buffer and force failure (e.g. invalid `--target-pages` via config).

### Repro C — Rust integration (optional)

From the repo, with `SKIP_INTEGRATION_TESTS` unset and ImageMagick unavailable:

```bash
cargo test -p liteparse test_parse_bytes_image_integration -- --nocapture
```

Then inspect the system temp directory for leftover `liteparse-*` folders after a failed run.

**Note:** This reproduces **leftover files on disk**, not necessarily a thrown error on the success path.

### Error Message

```text
no error is thrown on the success path. On failure paths, the parse/conversion error is returned as usual, but orphan temp directories remain under the OS temp path (e.g. /tmp/liteparse-* on Linux, %TEMP%\liteparse-* on Windows).
```

### LiteParse Version

2.0.0

### Operating System

Windows

### Node.js Version

v22.11.0

### Additional Context

_No response_

AdemBoukhris457:
Fixes #210 

## Summary

- Fixes a temp directory leak when parsing documents from in-memory bytes (`Buffer` / `Uint8Array` / `parse_bytes()`).
- Replaces `TempDir::keep()` and success-only manual cleanup with RAII via a new `BufferConversionTemps` guard that holds both the staging directory (original upload) and the conversion output directory.
- Ensures all LiteParse-created temp dirs are removed when `parse_input` exits, whether conversion or parsing succeeds or fails.

## Problem

Buffer uploads were written to a staging temp dir (`liteparse-staging-*`), but cleanup was unreliable:

- `convert_data_to_pdf` called `TempDir::keep()`, disabling automatic cleanup.
- Staging dirs were only removed manually on specific success paths.
- Parser cleanup at the end of `parse_input` only ran on success and could turn a successful parse into an error if cleanup failed.

This left orphaned copies of uploaded files in the system temp folder after conversion or parse failures — a disk, privacy, and server stability issue for API workloads.

## Solution

- Introduce `BufferConversionTemps { staging, output }` returned from `convert_data_to_pdf`.
- Hold the guard for the full duration of `parse_input` so both directories are dropped on any exit path.
- Remove the `is_temporary_path` parameter and manual `remove_dir_all` calls from `convert_to_pdf`.

### DIFF
--- a/crates/liteparse/src/conversion.rs
+++ b/crates/liteparse/src/conversion.rs
@@ -643,7 +643,9 @@ mod tests {
         assert!(
             !staging_path.exists(),
             "staging temp dir should be removed on drop"
-    
+        );
+    }
+
     // ── find_pdf_in_dir ──────────────────────────────────────────────────────
 
     /// Regression test for the LibreOffice filename-sanitisation bug.
--- a/crates/liteparse/src/conversion.rs
+++ b/crates/liteparse/src/conversion.rs
@@ -37,6 +37,16 @@ pub struct ConversionResult {
     pub original_extension: String,
 }
 
+/// Temp directories created during buffer-to-PDF conversion. Retained until
+/// dropped so staging and output dirs are removed on both success and failure.
+#[derive(Debug)]
+pub struct BufferConversionTemps {
+    #[allow(dead_code)]
+    staging: TempDir,
+    #[allow(dead_code)]
+    output: Option<TempDir>,
+}
+
 enum ConversionTool {
     LibreOffice,
     ImageMagick,
@@ -84,7 +94,6 @@ pub fn is_supported_extension(path: &str) -> bool {
 pub async fn convert_to_pdf(
     path: &str,
     password: Option<&str>,
-    is_temporary_path: bool,
 ) -> Result<(ConversionResult, Option<TempDir>), LiteParseError> {
     let ext = Path::new(path)
         .extension()
@@ -125,10 +134,6 @@ pub async fn convert_to_pdf(
             convert_office_document(path, tmp_dir.path().to_str().unwrap(), password).await?
         }
     };
-    // Remove temporary path for Office docs/images bytes
-    if is_temporary_path {
-        tokio::fs::remove_dir_all(Path::new(path).parent().unwrap()).await?;
-    }
 
     Ok((
         ConversionResult {
@@ -482,14 +487,24 @@ pub fn guess_extension_from_data(data: &[u8]) -> Option<String> {
 pub async fn convert_data_to_pdf(
     data: Vec<u8>,
     password: Option<&str>,
-) -> Result<(ConversionResult, Option<TempDir>), LiteParseError> {
+) -> Result<(ConversionResult, BufferConversionTemps), LiteParseError> {
     let ext = guess_extension_from_data(&data);
-    let tmp_dir = tempfile::Builder::new().prefix("liteparse-").tempdir()?;
-    let tmp_path = tmp_dir
-        .keep()
+    let staging_dir = tempfile::Builder::new()
+        .prefix("liteparse-staging-")
+        .tempdir()?;
+    let tmp_path = staging_dir
+        .path()
         .join(format!("input.{}", ext.unwrap_or("bin".to_string())));
     tokio::fs::write(&tmp_path, data).await?;
-    convert_to_pdf(tmp_path.to_str().unwrap(), password, true).await
+    let (converted, output_dir) =
+        convert_to_pdf(tmp_path.to_str().unwrap(), password).await?;
+    Ok((
+        converted,
+        BufferConversionTemps {
+            staging: staging_dir,
+            output: output_dir,
+        },
+    ))
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


logan-markewich:
This fix was kind of smelly. Simplified it a bit (no need to introduce a struct just to hold values). Also updated the test

logan-markewich (approved):

PR #221, opened 2026-05-24
fix: OCR embedded image regions selectively instead of full-page render

### LINKED ISSUE #220: [Bug] Any embedded image triggers full-page OCR instead of selective image-region OCR
### Description

LiteParse documents **selective OCR** — running OCR only where native text extraction is insufficient — but the Rust parser treats **any page with an embedded image** as requiring **full-page OCR**. That renders and OCRs the entire page bitmap even when the page already has plenty of native PDF text (e.g. a small logo, icon, or chart on an otherwise text-heavy slide).

Image bounding boxes are detected via PDFium, but they are used only as a **boolean trigger** (`has_images`). They are **not** used to crop or OCR individual image regions.

### Current behavior (v2 Rust)

In `crates/liteparse/src/ocr_merge.rs`, OCR need is decided per page:

```rust
let text_length: usize = page.text_items.iter().map(|item| item.text.len()).sum();
let page_obj = document.page((page.page_number - 1) as i32)?;
let has_images = !page_obj.image_bounds(25.0, 0.9).is_empty();

if text_length >= 100 && !has_images {
    continue;
}

// Full-page render + OCR
let bitmap = page_obj.render(dpi)?;
// … convert to RGB, send entire buffer to OcrEngine::recognize()
```

If the page has **≥ 100 characters** of native text **and no** embedded images → OCR is skipped.

If the page has **any** qualifying embedded image (≥ 25 pt, not a >90% page background) → **full-page** render at `config.dpi`, then OCR on the entire bitmap.

`image_bounds()` returns per-image boxes in viewport coordinates (`crates/pdfium/src/page.rs`), but those coordinates are never passed to OCR or used for cropping.

### Documented / expected behavior

- **`AGENTS.md`:** *"OCR only runs on embedded images where text extraction failed, not the entire document."*
- **`LiteParseConfig` doc comment** (`config.rs`): *"When true, runs on text-sparse pages and embedded images."* — implies targeted use of images, not whole-page OCR whenever an image exists.

Expected for a text-heavy page with a small logo:

1. Skip full-page OCR (native text is sufficient).
2. OCR **only the embedded image region(s)** where text may live inside the bitmap (e.g. scanned figure, text-in-image).

### Unused capability

PDFium wrapper already exposes per-image rendering:

```rust
// crates/pdfium/src/page.rs
pub fn render_image_object(&self, image_obj_index: usize) -> Result<Bitmap, PdfiumError>
```

This API is **never called** from the OCR pipeline. The live path always uses `page.render(dpi)` (full page).

### Secondary quality impact

After full-page OCR on text-heavy pages, merge uses a tight overlap filter (`overlaps_existing_text`, 2 pt tolerance). OCR words overlapping **any** native text bbox are dropped. On mixed pages, this can discard valid text from image regions while still paying full-page OCR cost.

### When full-page OCR is correct (not this bug)

- **Sparse pages** (`text_length < 100`): scanned or image-only pages — full-page OCR is appropriate.
- **`ocr_enabled: false`**: no OCR runs.

### Impact

| Area | Effect |
|------|--------|
| **Performance** | Full-page render + OCR per affected page; very slow on decks/reports where every slide has a logo |
| **Docs vs code** | Users and `AGENTS.md` describe selective OCR; code does "any image → whole page" |
| **Quality** | Full-page OCR + overlap dedup can miss text inside images on mixed pages |
| **HTTP OCR** | Full-page PNG encode per page (`http_simple.rs`) amplifies cost |

### Suggested fix direction

1. Split triggers:
   - `needs_full_page_ocr = text_length < 100`
   - `needs_image_ocr = has_images && text_length >= 100`
2. For image OCR on text-heavy pages:
   - Use `image_bounds()` results (already filtered: min 25 pt, max 90% coverage)
   - Render each image via `render_image_object()` or crop from a single page render
   - OCR each region; map bboxes back to page viewport space
3. Attach image bboxes to extracted `Page` during §2 extract to avoid re-scanning in OCR (optional perf win)
4. Filter merged OCR results to image regions on mixed pages (don't rely on overlap dedup alone)
5. Optional config `ocr_scope: 'selective' | 'full'` for backward compatibility

### Affected code

| Location | Behavior |
|----------|----------|
| `crates/liteparse/src/ocr_merge.rs` ~64–82 | Trigger logic + full-page `render(dpi)` |
| `crates/pdfium/src/page.rs` ~153–201 | `image_bounds()` — detection only |
| `crates/pdfium/src/page.rs` ~205–233 | `render_image_object()` — unused by OCR |
| `crates/liteparse/src/config.rs` ~8 | Doc comment implies selective image OCR |
| `AGENTS.md` §4 | Selective OCR design doc |

### Steps to Reproduce

1. Use a PDF where most pages have **lots of native text** (PowerPoint export, report, invoice) and **at least one small embedded image per page** (company logo, icon, figure).

2. Parse with OCR enabled (default):

   **CLI:**

   ```bash
   lit parse deck-with-logos.pdf --format json -q
   ```

   **Rust library:**

   ```rust
   let lp = LiteParse::new(LiteParseConfig { ocr_enabled: true, ..Default::default() });
   let result = lp.parse("deck-with-logos.pdf").await?;
   ```

   **Node.js:**

   ```typescript
   const parser = new LiteParse({ ocrEnabled: true, dpi: 150 });
   const result = await parser.parse("deck-with-logos.pdf");
   ```

3. Compare timing with `--no-ocr` on the same file — OCR-enabled run is much slower on every page with a qualifying embedded image, even when native text is abundant.

4. Observe stderr timing (when not `--quiet`):

   ```text
   [liteparse] extract: …ms (N pages)
   [liteparse] ocr: …ms        ← large on logo-heavy decks
   [liteparse] project: …ms
   ```

5. **Minimal repro:** One-page PDF with a paragraph of real text + one small embedded PNG logo (≥ 25 pt). Parse with default OCR — full page is rendered and sent to Tesseract/HTTP OCR despite `text_length >> 100`.

6. **Code-level confirmation (optional):** breakpoint or log in `ocr_merge.rs` after `page_obj.render(dpi)` — bitmap dimensions match full page at `dpi`, not image bbox size.

### Error Message

```text
no error is thrown. This is unexpected **behavior** (performance / incorrect OCR scope), not a crash.
```

### LiteParse Version

2.0.0

### Operating System

Windows

### Node.js Version

v22.11.0

### Additional Context

_No response_

AdemBoukhris457:
Closes #220 

## Summary
- Fixes OCR on text-heavy pages with embedded images: each qualifying image is rendered and OCR'd via `render_image_object()` instead of rendering and OCRing the entire page
- Sparse pages (`< 100` chars native text) still use full-page OCR (scanned / image-only pages)
- Maps OCR bboxes from image pixel space back to page viewport coordinates

## Motivation
Previously, any page with an embedded image (logo, icon, figure) triggered full-page OCR even when the page already had plenty of native text. `image_bounds()` was only used as a boolean trigger; image boxes were never used for cropping or regional OCR.

This caused unnecessary render + OCR cost on decks and reports where every slide has a small logo.

## Documentation alignment
Code now matches documented **selective OCR** behavior:
- **AGENTS.md:** *"OCR only runs on embedded images where text extraction failed, not the entire document."*
- **`LiteParseConfig`:** runs on text-sparse pages and embedded images — image OCR is now scoped to those regions, not the whole page

## Behavior change

| Page condition | Before | After |
|----------------|--------|--------|
| `< 100` chars | Full-page OCR | Full-page OCR (unchanged) |
| `≥ 100` chars + images | Full-page OCR | Per-image OCR only |
| `≥ 100` chars, no images | Skip | Skip (unchanged) |

## Test plan
- [x] `cargo test -p liteparse --no-default-features ocr_merge` (8/8 pass, incl. bbox mapping tests)
- [ ] Parse a text-heavy PDF with a small logo per page — confirm faster OCR vs previous full-page behavior
- [ ] Parse a scanned/sparse PDF — confirm full-page OCR still runs
- [ ] Parse same file with `--no-ocr` — unchanged

### DIFF
--- a/crates/liteparse/src/ocr_merge.rs
+++ b/crates/liteparse/src/ocr_merge.rs
@@ -4,7 +4,13 @@ use crate::error::LiteParseError;
 use crate::ocr::{OcrEngine, OcrOptions, OcrResult};
 use crate::types::{Page, TextItem};
 use image::{ImageBuffer, Rgba};
-use pdfium::Document;
+use pdfium::{Bitmap, Document};
+
+/// Minimum native text length before skipping full-page OCR.
+const SPARSE_PAGE_TEXT_THRESHOLD: usize = 100;
+/// Match `Page::image_bounds` filters in the OCR trigger path.
+const MIN_IMAGE_SIZE_PT: f32 = 25.0;
+const MAX_IMAGE_PAGE_COVERAGE: f32 = 0.9;
 
 /// Owned page bitmap prepared for OCR. Indices refer to positions in the `pages` slice.
 pub(crate) struct RenderedPage {
@@ -12,9 +18,26 @@ pub(crate) struct RenderedPage {
     pub rgb_bytes: Vec<u8>,
     pub width: u32,
     pub height: u32,
+    pub coord_map: OcrCoordMap,
+}
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum OcrCoordMap {
+    FullPage {
+        dpi: f32,
+    },
+    ImageRegion {
+        page_x: f32,
+        page_y: f32,
+        page_w: f32,
+        page_h: f32,
+    },
 }
 
 /// Render pages that need OCR from an already-open document.
+///
+/// OCR is triggered when a page has fewer than [`SPARSE_PAGE_TEXT_THRESHOLD`] characters
+/// of native text (full-page OCR) or has qualifying embedded images (per-image OCR).
 ///
 /// The pdfium `Document` holds raw pointers that are not `Send`, so callers must
 /// drop it before awaiting the OCR engine.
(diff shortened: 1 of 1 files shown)


logan-markewich:
I don't disagree with this change, but some complex/annoying merge conflicts popped up based on past PRs

AdemBoukhris457:
> I don't disagree with this change, but some complex/annoying merge conflicts popped up based on past PRs

@logan-markewich Just rebased on main and took care of the conflicts. Should be good to go now

adarshmadrecha:
Just noting a edge case scenario, Sometimes, the PDF creating software will split the image and them embed it into the PDF. 


Check this PDF for instance
[Uploading 02.pdf…]()

Interesting pages
 - 32 -- Here the Image is occupying almost half the page. In this case, I would suggest only picking up per image OCR
 - 39 -- Here the Image is occupying almost all the area of the page. Except for the Header which appears on all the pages, there is NO text. In this case, I would suggest Full-page OCR. Technically, there are 3 images on this page because it got split.

logan-markewich:
Going to close this for now, since its not an optimization I want to make at the moment

PR #244, opened 2026-05-28
feat: Add OCR-only text mode for corrupted native PDF text layers

patelvivekdev:
## Summary

Ports the OCR text-combination option from the pre-Rust PR #168 to the current Rust core and bindings.

Adds `ocrTextMode` / `ocr_text_mode` with two modes:

- `merge` (default): keep native PDF text and add non-overlapping OCR text, preserving existing behavior.
- `ocr-only`: on pages where OCR runs, use OCR-derived text items as the source of truth and replace native text items.

This is useful for PDFs with corrupted, hidden, or misleading native text layers where OCR reads the visible page correctly but merge mode would discard overlapping OCR results.

## Behavior

`ocr-only` only changes how text is combined after OCR has already been selected for a page. It does not force OCR to run on every page.

For OCR'd pages in `ocr-only` mode:

- OCR confidence filtering is preserved.
- Native-text overlap filtering is skipped.
- `page.text_items` is replaced with cleaned OCR text items.

## Updated surfaces

- Rust core config and OCR merge path
- Rust CLI: `--ocr-text-mode ocr-only`
- Node native config, TypeScript wrapper, CLI, and lockfile
- Python PyO3 config, Python wrapper, and CLI
- WASM config
- README docs

## Verification

Ran locally:

- `npm install --ignore-scripts` in `packages/node`
- `npm ci --ignore-scripts` in `packages/node`
- `npx tsc` in `packages/node`
- `cargo fmt --all -- --check`
- `cargo clippy --all-targets --no-default-features`
- `cargo test -p liteparse --no-default-features ocr_text_mode --lib`
- `cargo test -p liteparse --no-default-features ocr_results_to_text_items --lib`
- `cargo check -p liteparse --no-default-features`
- `cargo check -p liteparse-napi --no-default-features`
- `cargo check -p liteparse-python --no-default-features`
- `cargo check -p liteparse-wasm`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,9 @@ lit parse document.pdf --target-pages "1-5,10,15-20"
 # Parse without OCR
 lit parse document.pdf --no-ocr
 
+# Use OCR text as the source of truth on pages where OCR runs
+lit parse document.pdf --ocr-text-mode ocr-only
+
 # Parse a remote PDF
 curl -sL https://example.com/report.pdf | lit parse -
 ```
--- a/crates/liteparse-napi/src/types.rs
+++ b/crates/liteparse-napi/src/types.rs
@@ -1,6 +1,6 @@
 use napi_derive::napi;
 
-use liteparse::config::{LiteParseConfig, OutputFormat};
+use liteparse::config::{LiteParseConfig, OcrTextMode, OutputFormat};
 use liteparse::parser::ParseResult;
 use liteparse::types::{ParsedPage, TextItem};
 
@@ -17,6 +17,8 @@ pub struct JsLiteParseConfig {
     pub ocr_enabled: Option<bool>,
     /// HTTP OCR server URL. If set, uses HTTP OCR instead of Tesseract.
     pub ocr_server_url: Option<String>,
+    /// How OCR text is combined with native text: "merge" or "ocr-only".
+    pub ocr_text_mode: Option<String>,
     /// Path to tessdata directory for Tesseract.
     pub tessdata_path: Option<String>,
     /// Maximum number of pages to parse.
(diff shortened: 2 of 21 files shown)


logan-markewich:
Do you have an example doc that was causing the original issue (i.e. scanned doc that so.ehow wasnt triggering OCR)

Id honestly like to make the OCR detection more robust, like #236

logan-markewich:
Give 2.0.3 a try for this

logan-markewich:
Actually, I ran into some cases where this is useful. There are some pretty weird PDFs out there

patelvivekdev:
Yes, we have an example, but it's proprietary, so I can't share the results. 

We tried to run with that https://github.com/run-llama/liteparse/pull/168, and it was working.

patelvivekdev:
> Give 2.0.3 a try for this

Sure, I'll test it with the new version. Right now, still using an older version.

patelvivekdev:
@logan-markewich I've tested V2 with our PDFs.

✅  The speed has been great so far.

However, we've noticed some inconsistencies compared to V1.

Details: https://github.com/patelvivekdev/liteparse-v2-ocr-bug/blob/main/README.md

Background: We are using Azure DI for OCR. It worked well in V1, but now I'm encountering an issue in V2 where, after obtaining the correct OCR text items, Liteparse does not add them back.

logan-markewich:
@patelvivekdev yea probably similar to what #265 is reporting. Do you have a PDF or two with this symptom that I can validate against?

logan-markewich:
ah, I see your project generates PDFs, will take a look

patelvivekdev:
I have other examples that I'll add to the same repo today.

patelvivekdev:
One thing I want to clarify here is that we now have two issues.

1. Missing lines. Details: https://github.com/patelvivekdev/liteparse-v2-ocr-bug/blob/main/README.md

2. The actual bug: when a page has a garbled block and needs OCR, we prioritize the garbled text over the OCR result.

logan-markewich:
Both fixable easily. Will take a look, thanks

logan-markewich:
2.0.5 should fix both issues

PR #249, opened 2026-05-29
fix: pass DocumentOpenPassword filter option to LibreOffice for encrypted Office docs

### LINKED ISSUE #212: [Bug] --password is passed to LibreOffice as --infilter, breaking encrypted Office document conversion
### Description

LiteParse documents `--password` / `password` config as working for **encrypted/protected documents** across CLI and all bindings. For **PDFs**, the password is correctly forwarded to PDFium during load/extract. For **Office documents** (`.docx`, `.xlsx`, `.pptx`, etc.), the password is passed into LibreOffice conversion using the wrong CLI mechanism, so password-protected Office files fail to convert or behave unpredictably.

### Current (broken) implementation

In `crates/liteparse/src/conversion.rs`, `convert_office_document` does:

```rust
if let Some(pw) = password {
    infilter_arg = format!("--infilter=:{pw}");
    args.push(&infilter_arg);
}
```

Example command shape:

```bash
soffice -env:UserInstallation=file://... \
  --headless --invisible \
  --convert-to pdf --outdir /tmp/out \
  --infilter=:mySecretPassword \
  /path/to/protected.docx
```

### Why this is wrong

LibreOffice's `--infilter` flag specifies an **input filter type name**, not a document password. From the [soffice man page](https://manpages.ubuntu.com/manpages/noble/man1/soffice.1.html):

> `--infilter=" "` — Force an input filter type if possible. For example `--infilter="Calc Office Open XML"` …

Passing `--infilter=:mySecretPassword` tells LibreOffice to use a filter literally named `:mySecretPassword`, which is not a valid filter. LibreOffice has **no documented `--password` CLI flag** for headless conversion in the same way tools like `unoconv --password` provide.

### What works vs what doesn't

| Document type | Password path | Status |
|---------------|---------------|--------|
| Encrypted PDF (`.pdf`) | PDFium via `extract.rs` / `load_document(..., password)` | **Works** |
| Password-protected Office (`.docx`, `.xlsx`, `.ppt`, …) | LibreOffice via `convert_office_document` | **Broken** |
| Office buffer upload with password | Same broken LibreOffice path | **Broken** |

### User-facing impact

- Users following docs (`--password`, `password=` in Python/Node) expect it to unlock **any** supported encrypted document.
- Password-protected Office files fail conversion with errors like "output PDF not found" or opaque LibreOffice stderr — not a clear "wrong password" message.
- Silent misconfiguration: the password is never applied correctly, so users may think the file or LiteParse is broken.

### Expected behavior

When `password` is set and the input is a password-protected Office document:

1. LibreOffice should receive the password through a supported mechanism (e.g. filter options / UNO env vars / documented headless API).
2. Conversion should produce a PDF on success.
3. Wrong password should return a clear conversion error.

### Possible fix directions

- Use LibreOffice filter options for open password (e.g. `DocumentOpenPassword` via export/import filter properties — similar to how `unoconv -e DocumentOpenPassword=...` works).
- Pass password via supported `-env:` variables if validated against the target LibreOffice version.
- Split documentation: clarify that `--password` currently applies to **PDFs only** until Office conversion is fixed.
- Add integration test with a known password-protected `.docx` fixture.

### Affected code

| Location | Behavior |
|----------|----------|
| `crates/liteparse/src/conversion.rs` — `convert_office_document` | Builds `--infilter=:{password}` |
| `crates/liteparse/src/conversion.rs` — `convert_to_pdf` | Forwards `password` to office conversion |
| `crates/liteparse/src/parser.rs` | Passes `config.password` into conversion |
| `crates/liteparse/src/main.rs` | CLI `--password` for `parse`, `batch-parse`, `screenshot` |
| `packages/node/src/lib.ts`, `packages/python/liteparse/parser.py` | Expose `password` for "protected documents" |
| `README.md`, `docs/.../cli-reference.md` | Document `--password` for encrypted/protected documents generally |

### Steps to Reproduce

1. Install LiteParse and LibreOffice.

2. Create or obtain a **password-protected Office document** (e.g. `.docx` encrypted with password `test123`).

3. Attempt to parse with the documented password flag:

   **CLI:**
   ```bash
   lit parse protected.docx --password test123 --no-ocr -q
   ```

   **Python:**
   ```python
   from liteparse import LiteParse

   parser = LiteParse(ocr_enabled=False, password="test123")
   result = parser.parse("protected.docx")
   ```

   **Node.js:**
   ```typescript

   const parser = new LiteParse({ ocrEnabled: false, password: "test123" });
   await parser.parse("protected.docx");
   ```

4. Observe conversion failure (no output PDF, or LibreOffice error in stderr).

5. **Control — same password on encrypted PDF works:**
   ```bash
   lit parse protected.pdf --password test123 --no-ocr -q
   ```

6. **Inspect the LibreOffice command** (optional): add logging or run equivalent CLI manually — note `--infilter=:test123` is not a password argument.

### Error Message

```text

```

### LiteParse Version

2.0.0

### Operating System

Windows

### Node.js Version

node -v

### Additional Context

_No response_

fuleinist:
## Summary

Fixes run-llama/liteparse#212.

### The bug

`convert_office_document` previously built the `--infilter` argument as:

```
--infilter=:mySecretPassword
```

LibreOffice interprets `:mySecretPassword` as the **filter type name** (the colon is the separator between filter name and options). Since no filter is named `:mySecretPassword`, the argument is silently ignored and the password is never applied.

### The fix

Replace with the correct LibreOffice filter-option syntax:

```
--infilter=docx:DocumentOpenPassword=mySecretPassword
```

This passes the password as the `DocumentOpenPassword` option to the appropriate input filter. The filter name is derived from the file extension so the option is scoped correctly.

### Changed file

- `crates/liteparse/src/conversion.rs` — `convert_office_document` function

### DIFF
--- a/crates/liteparse/src/conversion.rs
+++ b/crates/liteparse/src/conversion.rs
@@ -427,7 +427,6 @@ pub async fn convert_office_document(
             ))
         })?;
     let user_profile_url = format!("-env:UserInstallation={user_profile_file_url}");
-    let infilter_arg;
     let mut args: Vec<&str> = vec![
         &user_profile_url,
         "--headless",
@@ -438,8 +437,34 @@ pub async fn convert_office_document(
         output_dir,
     ];
     if let Some(pw) = password {
-        infilter_arg = format!("--infilter=:{pw}");
-        args.push(&infilter_arg);
+        // Derive the LibreOffice filter name from the file extension so the
+        // DocumentOpenPassword option is associated with the correct input filter.
+        let ext = Path::new(file_path)
+            .extension()
+            .and_then(|e| e.to_str())
+            .unwrap_or("")
+            .to_lowercase();
+        let filter_name = match ext.as_str() {
+            "doc" | "docx" | "docm" | "dot" | "dotm" | "dotx" => "docx",
+            "xls" | "xlsx" | "xlsm" | "xlsb" => "xlsx",
+            "ppt" | "pptx" | "pptm" | "pot" | "potm" | "potx" => "pptx",
+            "odt" | "ott" => "odt",
+            "ods" | "ots" => "ods",
+            "odp" | "otp" => "odp",
+            "rtf" => "rtf",
+            "pages" => "pages",
+            _ => "",
+        };
+        if !filter_name.is_empty() {
+            // LibreOffice --infilter accepts "FilterName:option=value" to pass
+            // per-filter open options. DocumentOpenPassword is the standard option
+            // for supplying a document-open password (distinct from the
+            // write-protection password).  This replaces the previous broken
+            // approach of passing --infilter=:{pw} which treated the password
+            // itself as the filter type name.
+            let infilter_arg = format!("--infilter={filter_name}:DocumentOpenPassword={pw}");
+            args.push(&infilter_arg);
+        }
     }
     args.push(file_path);
 


crates/liteparse/src/conversion.rs
```diff
@@ -438,8 +437,34 @@ pub async fn convert_office_document(
...
+            _ => "",
+        };
+        if !filter_name.is_empty() {
+            // LibreOffice --infilter accepts "FilterName:option=value" to pass
+            // per-filter open options. DocumentOpenPassword is the standard option
+            // for supplying a document-open password (distinct from the
+            // write-protection password).  This replaces the previous broken
+            // approach of passing --infilter=:{pw} which treated the password
+            // itself as the filter type name.
+            let infilter_arg = format!("--infilter={filter_name}:DocumentOpenPassword={pw}");
```
logan-markewich, line 465: I... dont know if this is correct? Doesn't this add a password rather than decrypting for a conversion?

Have you tested this change? Or do you have a doc to test with that you can share?

fuleinist: my bad, closing this PR now for self retro 🙏

PR #276, opened 2026-06-05
ci: build Linux glibc binaries on ubuntu-22.04 for portability

jmho-govsignals:
The linux-x64-gnu / linux-arm64-gnu .node bindings and the lit cli build on Ubuntu 24.04 runners (ubuntu-latest / ubuntu-24.04-arm), whose glibc 2.39 / GCC 13 toolchain requires GLIBC_2.39 and GLIBCXX_3.4.31. This causes the package to fail to load on any older-glibc distro, e.g. Debian bookworm (glibc 2.36) with version 'GLIBCXX_3.4.31' not found.

This PR drops the runner to ubuntu-22.04 which increases the portability of the package since that runner lowers the binary's requirement to GLIBC_2.34 / GLIBCXX_3.4.29. It might make more sense to do a proper CI matrix builder strategy but this opts for least changes.

### DIFF
--- a/.github/workflows/release-node.yml
+++ b/.github/workflows/release-node.yml
@@ -53,12 +53,12 @@ jobs:
             node-arch: darwin-x64
             pdfium-dylib: libpdfium.dylib
 
-          - os: ubuntu-latest
+          - os: ubuntu-22.04
             target: x86_64-unknown-linux-gnu
             node-arch: linux-x64-gnu
             pdfium-dylib: libpdfium.so
 
-          - os: ubuntu-24.04-arm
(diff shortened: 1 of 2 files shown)


logan-markewich:
We would need this change for more than just node right? Any reference to ubuntu-latest should be replaced?

jmho-govsignals:
> We would need this change for more than just node right? Any reference to ubuntu-latest should be replaced?

Agreed. The python release was also impacted by this. I missed it since I was only using it on node but as far as I know the other builders should be covered

logan-markewich (approved):

PR #282, opened 2026-06-08
feat: make table cleaning opt-out by default

nnethercott:
# description
Depending on the ocr provider configured in the`HttpOcrEngine`, partial table extractions are possible/likely. For example, consider tables with multi-row cell merges which don't render nicely in markdown. 

Rather than discarding these partial extractions, many RAG pipelines would still benefit from the unstructured text. 

This PR simply adds a new argument to the `LiteParseConfig` (and corresponding bindings) which determines whether or not `clean_ocr_table_artifacts` gets invoked.

### DIFF
--- a/crates/liteparse-napi/src/types.rs
+++ b/crates/liteparse-napi/src/types.rs
@@ -35,6 +35,8 @@ pub struct JsLiteParseConfig {
     pub quiet: Option<bool>,
     /// Number of concurrent OCR workers (default: CPU cores - 1).
     pub num_workers: Option<u32>,
+    /// Whether to auto-remove invalid table artifacts from raw OCR response.
+    pub clean_artifacts: Option<bool>,
 }
 
 impl JsLiteParseConfig {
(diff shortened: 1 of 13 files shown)


logan-markewich:
@nnethercott do you have an example doc here? Or a before/after sample? I'm kinda curious what this is actually fixing

logan-markewich:
This isn't discarding tables, and there is no markdown. Its just stripping chars that tends to create messy OCR results on tables

nnethercott:
Hey @logan-markewich coming back to this now. I've added a unit test on my side and you're right - it's not this (i'll close the PR). 

The bug i'm having is actually when the ocr backend performs full-page ocr (so that the `OcrResult.bbox` == page dims) which means it gets skipped [here](https://github.com/run-llama/liteparse/blob/1147b883378ebd77488db0da9f4911054aa9624c/crates/liteparse/src/ocr_merge.rs#L200-L209) when any text item is found deterministically (not by the ocr engine) in that region.  

This makes sense for engines like paddle ocr which do return bboxes per span/paragraph, but when using something like [mistral ocr](https://docs.mistral.ai/api/endpoint/ocr) the response is just a blob of all text found for the page. 

Maybe an argument could instead be added to the engine for allowing overlaps ? wdyt

PR #285, opened 2026-06-08
Add opt-in inline base64 image extraction

saravananravi08:
## Summary

Adds an opt-in `inline_images` / `inlineImages` feature that extracts embedded PDF images, encodes them as PNG base64, and inserts them into parsed page text as markdown data URI images:

```md
![](data:image/png;base64,...)
```

This allows consumers to render parsed text and embedded images together from the returned text stream, while keeping the existing default behavior unchanged.

## Why this is useful

LiteParse currently does a strong job extracting text and coordinates, but many real-world PDFs contain important non-text content that gets lost in text-only output, such as:

- signatures, seals, stamps, and letterheads
- charts, diagrams, screenshots, and figures
- QR codes, barcodes, logos, and product images
- form attachments or image snippets embedded inside a mostly text-based PDF

Inlining images as markdown data URIs makes the parse result self-contained and directly renderable by downstream applications without needing separate image files, storage, or hosting.

## Where this helps

- **Web app document preview:** render `result.text` as markdown and show both text and images in order.
- **RAG / LLM ingestion:** preserve visual context that may be important for document understanding.
- **Invoice, form, and report parsing:** keep signatures, stamps, charts, product photos, QR codes, and similar embedded assets near surrounding text.
- **Markdown export:** save parsed output as one `.md` document with embedded images.
- **Offline/browser workflows:** avoid external image URLs because the image bytes are included directly in the parse result.

The feature also exposes structured image metadata on pages, so consumers can either use the inline markdown text or use coordinates/base64 for custom rendering.

## What changed

- Added embedded image extraction/rendering via PDFium image objects.
- Added `ImageItem` metadata on parsed pages.
- Added inline image placeholders before grid projection and replacement after projection.
- Ensured inline image markdown lines are left-trimmed so they are not treated as indented markdown/code blocks.
- Added `inline_images` / `inlineImages` config support across:
  - Rust core + CLI
  - Node/NAPI bindings + CLI
  - Python bindings + CLI
  - WASM bindings
- Added structured image output in JSON/bindings when inline images are enabled.
- Added implementation docs for inline base64 images and WASM OCR memory-safe rendering.

## Usage

Rust CLI:

```bash
lit parse document.pdf --inline-images
```

Node:

```ts
const parser = new LiteParse({ inlineImages: true });
const result = await parser.parse("document.pdf");
```

Python:

```py
parser = LiteParse(inline_images=True)
result = parser.parse("document.pdf")
```

## Notes

This feature is disabled by default because embedded base64 images can significantly increase output size and memory usage.

## Validation

Ran successfully:

```bash
cargo fmt
cargo check -p liteparse -p liteparse-napi -p liteparse-python --all-targets
cargo check -p liteparse-wasm
npm --prefix packages/node run build:ts
cargo test -p liteparse --lib
cargo test -p liteparse image_merge --lib
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1207,6 +1207,7 @@ checksum = "92daf443525c4cce67b150400bc2316076100ce0b3686209eb8cf3c31612e6f0"
 name = "liteparse"
 version = "2.1.1"
 dependencies = [
+ "base64",
  "clap",
  "image",
  "infer",
--- a/crates/liteparse-napi/src/types.rs
+++ b/crates/liteparse-napi/src/types.rs
@@ -44,6 +44,8 @@ pub struct JsLiteParseConfig {
     /// (default — emits `![](image_pN_K.png)` references with no bytes), or
     /// "embed" (also returns each image's PNG bytes on `images`).
     pub image_mode: Option<String>,
+    /// Inline embedded PDF images into markdown text as base64 data URI images.
+    pub inline_images: Option<bool>,
     /// Render hyperlink annotations as `[text](url)` in markdown output
     /// (default true). Set false for plain anchor text.
     pub extract_links: Option<bool>,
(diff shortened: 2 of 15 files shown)


saravananravi08:
Closing this since image mode is added in markdown after i added inline image in text, which satisfies my requirement

PR #286, opened 2026-06-09
feat: add docker-compose for EasyOCR server

quanhua92:
Makes it easy to run the EasyOCR server with a single command.

### Usage:

**Spin up the server:**
```bash
docker compose -f ocr/easyocr/docker-compose.yml up -d
```

**Test with curl:**
```bash
curl -X POST http://localhost:8828/ocr -F "file=@./integration_tests_data/receipt.png" -F "language=en"
```

**Test with lit:**
```bash
lit parse --ocr-server-url http://localhost:8828/ocr document.pdf
```
- Runs on ARM64 and amd64
- EasyOCR models cached in named Docker volume
- Health check on /health, restart unless stopped

### DIFF
--- a/ocr/easyocr/docker-compose.yml
+++ b/ocr/easyocr/docker-compose.yml
@@ -1,0 +1,19 @@
+services:
+  easyocr:
+    build: .
+    container_name: easyocr
+    ports:
+      - "8828:8828"
+    volumes:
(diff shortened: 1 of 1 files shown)


logan-markewich:
Just going to close this for now to keep the example straightforward (if we add docker compose, would rather keep all examples aligned)

PR #288, opened 2026-06-09
Add opt-in vector chart ASCII detection

saravananravi08:
## Summary

Adds an opt-in chart detection path for simple vector bar charts in PDFs. When enabled, LiteParse renders pages at a lightweight chart-detection DPI, detects grouped vertical bar charts from visual components, infers labels/scale from extracted text items, and replaces the text-only chart outline with an ASCII bar chart.

## Why this is useful

Some PDFs draw charts as vector paths instead of embedded images. Existing text extraction can capture axis/legend labels, but it misses the visual bar values because there is no raster image object to extract. This gives users a readable text representation for supported vector bar charts without requiring OCR or image extraction.

## Behavior

- Disabled by default
- Enabled with `--detect-charts`
- Supports simple vertical/grouped bar charts in v1
- Uses `█` block bars for ASCII output
- Does not insert a `[Detected ...]` heading into page text
- Leaves output unchanged silently for unsupported or low-confidence detections
- Exposes structured chart metadata in JSON/Node/WASM result pages

## API / CLI

Rust CLI:

```bash
lit parse file.pdf --detect-charts
lit parse file.pdf --detect-charts --chart-detection-dpi 100
```

Node:

```ts
const parser = new LiteParse({ detectCharts: true });
```

Python:

```py
parser = LiteParse(detect_charts=True)
```

## Validation

Ran:

```bash
cargo test -p liteparse chart --lib
cargo check -p liteparse -p liteparse-napi -p liteparse-python --all-targets
cargo check -p liteparse-wasm
npm --prefix packages/node run build:ts
```

Manual test with `image-inline-test.pdf` page 1 confirmed:

- default parse remains unchanged
- `--detect-charts` replaces the vector chart outline with ASCII bars
- no detection heading is inserted into page text
- JSON output includes one structured chart item

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1207,6 +1207,7 @@ checksum = "92daf443525c4cce67b150400bc2316076100ce0b3686209eb8cf3c31612e6f0"
 name = "liteparse"
 version = "2.0.6"
 dependencies = [
+ "base64",
  "clap",
  "image",
  "infer",
--- a/README.md
+++ b/README.md
@@ -201,6 +201,7 @@ Options:
       --preserve-small-text    Keep very small text
       --password <password>    Password for encrypted documents
       --num-workers <n>        Concurrent OCR workers [default: CPU cores - 1]
+      --inline-images          Inline embedded PDF images as base64 markdown images
   -q, --quiet                  Suppress progress output
   -h, --help                   Print help
 ```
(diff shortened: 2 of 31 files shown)


logan-markewich:
This is neat, thanks! But I think I'm going to decline this for now. Full-page screenshots exist to cover this gap, and this implementation feels both too narrow to be useful while contributing a decent amount of code for such a niche feature.

Miss-detection has a huge cost since we are deleting lines of text based on heuristic matches. I'm also not convinced that the current output format is helpful compared to just fetching the full-page screenshot.

I'm also super hesitant to merge in more knobs for niche features (its nice when things "just work" as much as possible)

PR #306, opened 2026-06-18
Avoid wasm render crashes from PDFium-owned bitmap buffers

YageGeng:
## Summary
- On `wasm32`, allocate the `Bitmap::new()` BGRA backing store in Rust and pass it to `FPDFBitmap_CreateEx` with an explicit stride.
- Keep that Rust-owned buffer alive on the `Bitmap` wrapper and return it from `Bitmap::buffer()` for wasm-created bitmaps.
- Reject non-positive bitmap dimensions before entering PDFium.

## Why
This is intended to make browser/Wasm page rendering robust. The page render path (`Page::render()` → `Bitmap::new()` → `FPDF_RenderPageBitmap()` → `to_rgba()` / `to_rgb()`) currently asks PDFium to allocate the bitmap internally by passing a null external buffer to `FPDFBitmap_CreateEx`. In the wasm build, PDFium is linked as a static WASI archive while Rust exposes the rendered pixels as Rust slices. That leaves the bitmap backing memory owned by PDFium across the wasm/PDFium boundary, which is fragile for browser rendering and can surface as crashes or invalid pixel reads when rendered page images are read back.

Passing an explicit Rust-owned buffer gives the wrapper clear ownership of the pixel memory, keeps it alive for the `Bitmap` lifetime, fixes the stride, and makes the bytes returned by `Bitmap::buffer()` stable for Rust-side image conversion. Without this, wasm rendering depends on PDFium-owned bitmap allocation semantics that are not safe to expose as a Rust slice in every wasm runtime.

## Reproduction / validation notes
I added a temporary diagnostic export locally that loads `integration_tests_data/sample.pdf`, calls `Page::render()` in the wasm package, and reads the bitmap through `to_rgba()`. In local Chrome headless on this machine, upstream `main` did not crash for that sample at 72/150/300/600 DPI, so I am not claiming this specific sample reproduces the crash. The diagnostic still exercises the exact render/readback path this patch changes.

## Test plan
- `cargo fmt --check`
- `git diff --check`
- `cargo check -p liteparse-pdfium`
- `cargo check -p liteparse-pdfium --target wasm32-unknown-unknown`
- `cargo test -p liteparse-pdfium`
- Local diagnostic browser check of `Page::render()` + `to_rgba()` via wasm web target at 72/150/300/600 DPI

### DIFF
--- a/crates/pdfium/src/bitmap.rs
+++ b/crates/pdfium/src/bitmap.rs
@@ -11,6 +11,8 @@ use crate::library::Library;
 /// section.
 pub struct Bitmap<'lib> {
     handle: pdfium_sys::FPDF_BITMAP,
+    #[cfg(target_arch = "wasm32")]
+    buffer: Option<Vec<u8>>,
     _lib: PhantomData<&'lib Library>,
 }
 
@@ -23,6 +25,8 @@ impl<'lib> Bitmap<'lib> {
     pub unsafe fn from_handle(handle: pdfium_sys::FPDF_BITMAP) -> Self {
         Bitmap {
             handle,
+            #[cfg(target_arch = "wasm32")]
+            buffer: None,
             _lib: PhantomData,
         }
     }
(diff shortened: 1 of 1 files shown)


YageGeng:
Closing for now because the current local reproduction does not show a wasm render failure when this compatibility change is removed. We can reopen or submit a new PR if we get a concrete failing PDF/browser/runtime case.

PR #319, opened 2026-06-23
perf: OCR page bitmaps in grayscale for engines that binarize internally

jmho-govsignals:
Ran into some additional OOM issues using liteparse with tesseract so decided to optimize this code path.

One finding we had was tesseract binarizes images internally (Leptonica/Otsu on luma) and converts any RGB input to grayscale first. Handing it a grayscale buffer is equivalent input at 1 byte/px instead of RGB's 3. Doing this cuts the per-page bitmap retained through the OCR pass by ~3× (I measured ~35–44% lower peak RSS on large image-heavy PDFs).

This was scoped behind a new OcrEngine::prefers_grayscale (Tesseract opts in while default is off) so color-trained engines like PaddleOCR/EasyOCR keep receiving RGB and are unaffected. 

A natural follow-up is a config option so custom/HTTP engines can opt into grayscale too but wanted to get feedback on this before proceeding in that direction.


logan-markewich:
I think this direction seems pretty nice.

I wonder if instead of adding a config option to control grayscale, if HTTP providers should be able to declare their capabilities somehow? Like a `/capabilities` API in our expected protocol?

jmho-govsignals:
Ooo interesting. I'm about 50/50 on it. The win I see is that for OCR servers you don't own, `/capabilities` would make syncing painless. The tradeoff is complexity as you can probably get the same result with less code via a config option, especially since the HTTP server is currently a bring-your-own for custom OCR solutions, so you usually own both ends anyway. It's worth noting you'd might want that config default regardless as it's also the natural fallback when `/capabilities` is absent from the http server. Thoughts?

logan-markewich:
yea true, maybe a CLI option is best, but yea, probably lower priority until someone actually needs it

logan-markewich (approved):

PR #320, opened 2026-06-24
feat(render): per-page raster-image coverage

wey-gu:
`image_bounds` already computes per-page image-object bounds, but it only prints
them as JSON to stdout, so there is no way to get that data from the library. I
ran into this trying to tell scanned / image-only PDF pages apart from
born-digital text pages: a page with one large image and almost no extractable
text is probably a scan, while a text page with a small inline figure is not.

This adds `render::page_image_coverage(input, min_image_size_pt,
max_image_page_coverage, password)`, which returns a `Vec<PageImageCoverage>`
(per page: image block count, total coverage ratio, largest-image ratio). It
reuses the same `Page::image_bounds` filter the CLI uses, so the counted set
matches. The two thresholds the CLI had hardcoded (25pt min, 0.9 max) are now
shared `DEFAULT_MIN_IMAGE_SIZE_PT` / `DEFAULT_MAX_IMAGE_PAGE_COVERAGE` consts,
and `image_bounds` uses them too.

The call is read-only and returns one entry per page in page order. Added unit
tests for serialization and the missing-file error path, matching the existing
render tests.

Happy to wire this through the CLI and the Node / Python / WASM bindings if you
want it exposed there too. I kept this PR to the core library function.

### DIFF
--- a/crates/liteparse/src/render.rs
+++ b/crates/liteparse/src/render.rs
@@ -100,6 +100,75 @@ struct ImageBoundsOutput {
     height: f32,
 }
 
+/// Default minimum side length (in points) for a raster image object to be
+/// counted; smaller objects such as rule lines, bullets, and icons are ignored.
+pub const DEFAULT_MIN_IMAGE_SIZE_PT: f32 = 25.0;
+
+/// Default upper bound on a single image's page coverage; an object at or above
+/// this fraction of the page is treated as a full-page background and ignored.
+pub const DEFAULT_MAX_IMAGE_PAGE_COVERAGE: f32 = 0.9;
+
+/// Raster-image coverage for one page.
+#[derive(Debug, Clone, Serialize)]
+pub struct PageImageCoverage {
+    /// 1-based page number.
+    pub page_num: u32,
+    /// Number of raster image objects counted on the page (after filtering).
+    pub image_block_count: usize,
+    /// Combined area of those images over the page area, clamped to 1.0.
+    pub image_coverage: f32,
+    /// Area of the single largest image over the page area, clamped to 1.0.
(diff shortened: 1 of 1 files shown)


logan-markewich:
Thanks! Will take a look at this

ooc @wey-gu it sounds like #318 might be interesting to you (and curious about feedback there). Sounds like you are tackling something similar given this PR :)

wey-gu:
> Thanks! Will take a look at this
> 
> ooc @wey-gu it sounds like #318 might be interesting to you (and curious about feedback there). Sounds like you are tackling something similar given this PR :)


Ha, this is exactly the road I was heading down, sorry i missed #318 where you already had it in flight. `is_complex` is a cleaner cut than mine: I was only exposing raw per-page image coverage and planning to build the “does this page need OCR” decision on top myself. Closing mine (#320) in favor of this XD

One small thing from my side: would you be open to also surfacing the raw image-coverage ratio, not just has_substantial_images? In our pipeline we sometimes tune the threshold per source (scanned receipts and slide decks behave pretty differently), so having the float would let us route without re-deriving it. Totally fine if you’d rather keep the API to the boolean decision; needs_ocr already covers the common case.

`uncovered_vector_area` and `is_garbled` are nice touches by the way. Looking forward to using this!!!

i loved liteparse soooo much :p

logan-markewich:
Great feedback! Ill add that to the other PR 👍🏻  should make it for a release tomorrow 

Thanks for using liteparse 🚀

wey-gu:
> Great feedback! Ill add that to the other PR 👍🏻 should make it for a release tomorrow
> 
> Thanks for using liteparse 🚀

wow Thanks Logan!
Looking forward to the release!

Issue #322, opened 2026-06-24
[Feature] Individual words in text_items


osthues-labs:
### Problem Statement

We have use cases where it might be important to know where exactly a words starts and where it ends. Having whole sentences in the items prevents us from using this library.  

### Proposed Solution

Maybe there could be a parameter that configures the detail level for the text items and allows it to select words.

osthues-labs:
@Ekanshiitk I am not a contributor in the project and don't have permissions to set an assignee. I would super appreciate it though if you could pick this up :)

PR #324, opened 2026-06-25
adding the word_granularity feature

itsamine27:
this solve the issue #322

### DIFF
--- a/crates/liteparse/src/projection.rs
+++ b/crates/liteparse/src/projection.rs
@@ -2803,7 +2803,38 @@ fn clean_rendered_text(text: &str) -> String {
         .join("\n")
 }
 
-pub fn project_pages_to_grid(pages: Vec<Page>) -> Vec<ParsedPage> {
+pub fn project_pages_to_grid(pages: Vec<Page>, word_granularity: bool) -> Vec<ParsedPage> {
+    if word_granularity {
+        // Short-circuit: Map directly to ParsedPage without running XY-cut layout logic
+        return pages
+            .into_iter()
+            .map(|page| {
+                // Build a raw sequence string from your single word tokens
+                let combined_text = page
+                    .text_items
+                    .iter()
+                    .map(|it| it.text.as_str())
+                    .collect::<Vec<_>>()
+                    .join(" ");
+
+                ParsedPage {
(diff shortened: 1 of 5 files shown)


crates/liteparse/src/projection.rs
```diff
@@ -2803,7 +2803,38 @@ fn clean_rendered_text(text: &str) -> String {
         .join("\n")
 }
 
-pub fn project_pages_to_grid(pages: Vec<Page>) -> Vec<ParsedPage> {
+pub fn project_pages_to_grid(pages: Vec<Page>, word_granularity: bool) -> Vec<ParsedPage> {
+    if word_granularity {
+        // Short-circuit: Map directly to ParsedPage without running XY-cut layout logic
```
logan-markewich, line 2808: If we don't run XY-cut and grid projection, the markdown output and text output is ~unusable right?

itsamine27: the word_granularity flag is explicitly added for downstream consumer tasks that bypass standard text altogethe specifically bounding-box coordinate tracking and search-indexing backends.When a user requests word_granularity, they don't care about a readable markdown document they need the raw TextItem array containing isolated word tokens with their exact spatial $X/Y$ bounds intact skipping XY-cut is a performance optimization that prevents the layout engine from wastefully rebuilding sentences only for the downstream client to immediately split them back apart

itsamine27: @logan-markewich can review this PR plz and tell me what to change or what to add (its is working greatly now)

logan-markewich: @itsamine27 I didn't want to compromise the text/markdown output just to output word level bounding boxes. Did a slightly different approach in #334

itsamine27: Ah that makes total sense I see what you mean now. My approach was focusing purely on maximizing performance for raw coordinate extraction by bypassing the layout calculations entirely but preserving both outputs simultaneously is definitely the cleaner approach for the library's core API since #334 handles the layout integration while keeping the markdown output intact.
 I'm happy to close this out in favor of that implementation
 Thanks for the guidance and the review

PR #323, opened 2026-06-25
feat: Add Surya OCR 2 server (ocr/suryaocr)

wuxuedaifu:
## Summary

Adds `ocr/suryaocr/`, a FastAPI OCR server wrapping [Surya OCR 2](https://github.com/datalab-to/surya) (`surya-ocr` 0.20.0), a multilingual OCR foundation model. It conforms to the LiteParse OCR API spec (`OCR_API_SPEC.md`) and matches the conventions of the existing `ocr/easyocr/` and `ocr/paddleocr/` servers, so it works as a drop-in `--ocr-server-url` target.

## Details

- `POST /ocr` (multipart `file` + optional `language`) and `GET /health`, on port **8830** (easyocr 8828, paddleocr 8829).
- **Block-level output**: one `RecognitionPredictor` call per page; each layout block's HTML is stripped to plain text and mapped to `{text, bbox, confidence, polygon?}`. Surya 2 has no per-line text API, so block granularity is the native single-call path.
- The `language` field is accepted for API compatibility but unused — Surya 2 is multilingual and needs no per-language model. Device is auto-detected (CUDA/MPS/CPU), overridable via `TORCH_DEVICE`.
- Defensive result parsing (handles dict/attribute blocks, missing polygons, derives bbox from polygon, defaults confidence to 1.0).
- Ships `server.py`, `test_server.py`, `pyproject.toml`, `uv.lock`, `Dockerfile`, `README.md`.
- Wires the server into `ocr/README.md` (comparison table, recommendations, Available Servers, Quick Start) and `OCR_API_SPEC.md` (example implementations).

## Notes

- `pillow` is pinned `<11` and dev `httpx` `<0.28` to satisfy `surya-ocr` 0.20.0's dependency constraints.
- The `Dockerfile` copies `uv.lock` and uses `uv sync --frozen` for reproducible images (a small improvement over the sibling servers' Dockerfiles).

## Testing

- `uv run pytest test_server.py` — **9/9 passing**. Tests mock the Surya predictor, so they run without downloading any model.
- Manually smoke-tested against the real `surya-ocr` model: `POST /ocr` returns spec-conformant JSON (recognized text, correct `bbox`/`polygon`/`confidence`), with CPU fallback when no GPU is present.

PR #327, opened 2026-06-26
fix(suryaocr): run Surya 2 on GPU in a single llama.cpp container

wuxuedaifu:
## Summary

Makes the `ocr/suryaocr` server actually serve OCR. The merged Surya server (#323) shipped a **CPU-only image with no inference-backend configuration**, so a live `POST /ocr` had no VLM backend to talk to and returned empty results. This PR wires up Surya 2's llama.cpp backend end-to-end inside a **single GPU container**.

## What changed

**`Dockerfile` — single GPU image**
- Rebased on `ghcr.io/ggml-org/llama.cpp:server-cuda`, which ships a `qwen35`-capable `llama-server` (Surya 2's custom GGUF architecture — stock Ollama's bundled llama.cpp cannot load it), `libggml-cuda`, and the CUDA runtime.
- Surya spawns `llama-server` from `PATH` **inside the same container**, so one image runs both the API and the model server. Run with `--gpus all`.
- Adds `uv`, the Python/torch/Pillow runtime libs, a GGUF cache dir (`HF_HOME=/models`), and clears the base image entrypoint.

**`server.py` — backend selection + grammar fix**
- Selects the `llamacpp` backend before importing `surya`, and documents the `vllm` / `SURYA_INFERENCE_URL` overrides.
- Sets `SURYA_GUIDED_LAYOUT=false`. Surya's layout step is the only request that sends a grammar, and its regex `pattern` (`^\d{1,4} \d{1,4} \d{1,4} \d{1,4}$`) breaks the upstream llama.cpp json-schema→GBNF converter (`failed to parse grammar`) — which was the direct cause of the empty results. With it off, layout runs as free generation (Surya parses the output itself); block and full-page OCR never use a grammar.

## Why this was empty before

`RecognitionPredictor([image])` runs full-page OCR → output loops → falls back to layout + block OCR. The layout request sent the json-schema grammar → `llama-server` returned `400 failed to parse grammar` → no layout boxes → no blocks → `{"results":[]}`. Block OCR (the step that reads text) sends no grammar and works fine once it receives boxes.

## Testing

Verified on a 4× A100 box:
- Build succeeds; single container started with `--gpus all`.
- `GET /health` → `{"status":"healthy"}`.
- `POST /ocr` on a receipt image → **10 text blocks** with bboxes and confidence (e.g. `SHOP`, `Article Count Amount Tax`, `T-Shirt XL.....1 $14,95 20%`, `GROSS $75,80`), `llama-server` running on GPU.
- `9/9` unit tests pass (`pytest test_server.py`).

No change to the LiteParse OCR API contract — `/ocr` still returns `{ results: [{ text, bbox, confidence, polygon? }] }`.

### DIFF
--- a/ocr/suryaocr/Dockerfile
+++ b/ocr/suryaocr/Dockerfile
@@ -1,25 +1,51 @@
-FROM ghcr.io/astral-sh/uv:python3.12-trixie
+# Single image: GPU-enabled llama.cpp `llama-server` + the Surya OCR API.
+#
+# Based on the official CUDA llama.cpp server image, which already provides a
+# `qwen35`-capable `llama-server` (with libggml-cuda), the CUDA runtime, and
+# Python 3.12. Surya's llamacpp backend finds `llama-server` on PATH and spawns
+# it INSIDE this same container, so one image runs both the API and the model
+# server. Run with `--gpus all`.
+#
+# (Stock Ollama's bundled llama.cpp cannot load Surya 2's custom qwen35 GGUF;
+#  this upstream build can.)
+FROM ghcr.io/ggml-org/llama.cpp:server-cuda
 
-# System libraries needed by torch / Pillow / Surya
-RUN apt-get update && apt-get install -y \
-    libgomp1 \
-    libglib2.0-0 \
-    libsm6 \
-    libxext6 \
-    libxrender-dev \
+# Bring uv in from its official image (no external install script needed).
+COPY --from=ghcr.io/astral-sh/uv:0.11.21 /uv /usr/local/bin/uv
+
+# Python/runtime libs for Surya, torch, and Pillow.
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    python3-venv \
     libgl1 \
+    libglib2.0-0 \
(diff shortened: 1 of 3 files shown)


wuxuedaifu:
✅ **GPU test passed** — verified on a 4× A100 box, single container, `--gpus all`:

- `docker build -t suryaocr-liteparse:cuda .` succeeds
- `GET /health` → `{"status":"healthy"}`
- `POST /ocr` on a receipt image → **10 text blocks** with bboxes + confidence (e.g. `SHOP`, `Article Count Amount Tax`, `T-Shirt XL.....1 $14,95 20%`, `GROSS $75,80`), returned in ~85s on first call (GGUF download + `llama-server` spawn); `llama-server` running on GPU inside the same container
- `9/9` unit tests pass (`pytest test_server.py`)

The `SURYA_GUIDED_LAYOUT=false` change is what fixed the previously empty `{"results":[]}` — the layout grammar no longer trips llama.cpp's json-schema→GBNF converter.

logan-markewich (approved):

PR #340, opened 2026-07-04
fix: reflowing text sometimes squished words together

logan-markewich:
This fixes a few edge cases where words are incorrectly merged together


logan-markewich:
dataset diff shows improvements across the board, nice!

PR #352, opened 2026-07-14
feat(ocr): add optional oar-ocr backend

wey-gu:
## What

Adds a non-default, native-only `oar-ocr` feature with `OarOcrEngine`, using LiteParse's existing `OcrEngine` trait.

The adapter accepts a caller-configured `OAROCRBuilder` or built runtime and maps text, confidence, bounds, and polygons into LiteParse's existing result types. The recommended constructor uses batch size one and serializes page inference to avoid multiplying inference memory across workers.

LiteParse does not select or download models. `OcrOptions::language` is intentionally ignored because the recognition model and dictionary define language coverage. Tesseract and all default/CLI/binding behavior remain unchanged.

Companion proposal: #353.

## Verification

- feature tests: 240 passed
- unchanged-default tests: 244 passed
- current defaults + `oar-ocr` check, clippy, fmt
- live macOS arm64 smoke with caller-supplied PP-OCRv6-small models: 39 regions with polygons

This stays draft until maintainers decide whether the optional ONNX dependency graph, Rust 1.95 requirement, and feature-specific CI cost belong upstream. End-to-end document memory is not claimed to be bounded; LiteParse still pre-renders selected pages.

### DIFF
--- a/crates/liteparse/README.md
+++ b/crates/liteparse/README.md
@@ -140,9 +140,62 @@ let parser = LiteParse::new(LiteParseConfig::default())
     .with_ocr_engine(Arc::new(my_engine));
 ```
 
+For a native ONNX backend, enable `oar-ocr` and supply a detection model,
+recognition model, and matching character dictionary:
+
+```rust,no_run
+use liteparse::ocr::oar::OarOcrEngine;
+use liteparse::{LiteParse, LiteParseConfig};
+use std::path::Path;
+use std::sync::Arc;
+
+let models = Path::new("models");
+let engine = OarOcrEngine::from_models(
+    models.join("pp-ocrv6_small_det.onnx"),
+    models.join("pp-ocrv6_small_rec.onnx"),
+    models.join("ppocrv6_dict.txt"),
+)?;
+
+let parser = LiteParse::new(LiteParseConfig::default())
+    .with_ocr_engine(Arc::new(engine));
+# Ok::<(), Box<dyn std::error::Error>>(())
+```
+
+For opt-in model downloads, enable `oar-ocr-auto-download` and pass registered
+bare file names to the same constructor. On first use, `oar-ocr` downloads the
+files from ModelScope, verifies their SHA-256 digests, and caches them under
+`$OAR_HOME` (default `~/.oar`):
+
+```rust,no_run
+use liteparse::ocr::oar::OarOcrEngine;
+
+let engine = OarOcrEngine::from_models(
+    "pp-ocrv6_small_det.onnx",
+    "pp-ocrv6_small_rec.onnx",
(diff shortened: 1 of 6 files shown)


wey-gu:
Companion proposal: #353. It frames the maintainer decision, memory/MSRV/dependency trade-offs, explicit non-goals, and acceptance gates for moving this draft toward review.

logan-markewich:
This is a pretty neat feature honestly, I didn't know this crate existed 

I think I'm ~ok with this, but a few implications I noticed
- I probably would not enable this for bindings. Bindings don't have a nice way to opt into features like pure rust, and enabling this by default for bindings would bloat the binary size. This is subject to me actually checking the binary size lol
- we may or may not want pass-through features for oar-ocr? Right now it just uses default features, which means no cuda support etc. without digging into oar-ocr and enabling it in your own rust crate

Excited to benchmark latency vs tesseract now. I know the newest paddle models are very tiny, I would bet they almost beat tesseract

crates/liteparse/README.md
```diff
@@ -140,9 +140,34 @@ let parser = LiteParse::new(LiteParseConfig::default())
...
 ```
 
+For a native ONNX backend, enable `oar-ocr` and supply the models explicitly:
+
+```rust,no_run
+use liteparse::ocr::oar::{OAROCRBuilder, OarOcrEngine};
+use liteparse::{LiteParse, LiteParseConfig};
+use std::sync::Arc;
+
+let engine = OarOcrEngine::from_builder(OAROCRBuilder::new(
```
logan-markewich: This might be a limitation of oar-ocr, but it'd be nice to just pass a folder and/or model name?

logan-markewich: Does this download models at runtime or it only works if you pre-download the model?

wey-gu: fixed!

wey-gu:
Thanks @logan-markewich !

Kindly check the shape now?

Also, made some small bench in https://github.com/run-llama/liteparse/issues/353#issuecomment-4976804050 :D

Cheers//Wey

logan-markewich:
I'm honestly tempted to make paddleocrv6 tiny the default and remove tesseract lol

Tesseract quality is typically quite bad. And the speedup seems nice. But will dig a bit more into specifics

logxdx:
> I'm honestly tempted to make paddleocrv6 tiny the default and remove tesseract lol
> 
> Tesseract quality is typically quite bad. And the speedup seems nice. But will dig a bit more into specifics

Any plans on doing so? 
I am working on ocr and layout detection myself and was tempted on doing this too. There's also [this paddle-ocr-rs](https://github.com/mg-chao/paddle-ocr-rs)

PR #358, opened 2026-07-16
Expose failed OCR page numbers in parse results

riz-adnan:
## Summary
- Track OCR failures by 1-based page number
- Return failed OCR pages in Rust, Python, Node, WASM, and JSON output
- Include failed page numbers in fatal OCR errors

## Testing
- python -m compileall packages/python/liteparse
- cargo fmt --all
- cargo test ...


logan-markewich:
Failed results are supported with a `continue_on_error` option now, that should surface failed pages in the JSON

PR #364, opened 2026-07-19
feat(ocr): add optional oar-ocr backend

logan-markewich:
All credits to https://github.com/run-llama/liteparse/pull/352 (I couldn't push to that PR sadly, just added a few small nits)


wey-gu:
Sorry @logan-markewich i just realized it was not with allow maintainer editing checked

Thanks!

PR #371, opened 2026-07-21
Add PDF embedded image extraction parity

hexapode:
## Summary

- add explicit opt-in image extraction via extract_images / extractImages / --extract-images, defaulting to false
- preserve compatibility: ImageMode::Embed or image_output_dir also enables extraction
- keep default placeholder-only parsing lightweight without loading pixel metadata, filters, matrices, or image streams
- expose embedded-image bbox, intrinsic dimensions, rotation, format, name/path, duplicate relationship, and extraction error counts across Rust, Node, Python, and WASM
- preserve valid direct DCT JPEG streams losslessly and fall back to rendered PNG for other image sources
- add exact duplicate reuse using raw bytes plus image metadata, including canonical on-disk path reuse
- support configurable embedded-image output directories; CLI JSON remains metadata-only and never serializes bytes/base64
- fix filtered image-object indexing and use the extracted format in Markdown references

## Validation

- cargo fmt --all -- --check
- cargo check for core, NAPI, and Python with no default features/offline
- cargo test -p liteparse --lib --no-default-features --offline (238 passed)
- cargo test -p liteparse-pdfium --offline
- Python compileall
- git diff --check
- runtime single-parse PDFium/CLI smoke with --extract-images and --image-output-dir
- Rust batch CLI help confirms --extract-images

## Validation limitations

Default-feature checks were blocked because tesseract-rs writes outside the sandbox. The WASM check was blocked when Cargo attempted to unpack a cached dependency outside the sandbox. TypeScript dependency installation was rejected and node_modules were unavailable; Rust NAPI checks pass.

## Representation notes

Valid DCT JPEG streams are preserved. Non-DCT and JPX inputs currently fall back to PNG rather than optional lossy JPEG recompression. In-memory bytes remain available for backward compatibility, but JSON omits them. A path is populated only when image_output_dir is configured.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,9 @@ lit parse document.pdf --format markdown --image-mode off
 # Extract embedded images to disk and reference them from the markdown
 lit parse document.pdf --format markdown --image-mode embed --image-output-dir ./images
 
+# Extract image bytes and metadata without changing Markdown image handling
+lit parse document.pdf --format json --extract-images
+
 # Emit link text as plain text (no [text](url) syntax)
 lit parse document.pdf --format markdown --no-links
 ```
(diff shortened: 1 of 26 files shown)


logan-markewich:
- Default behavior changes hiding inside: image bboxes now go through bounds_to_viewport instead of page_height - top — arguably a correctness fix, but it shifts figure placement on CropBox-offset pages in default output. And image_output_dir alone now silently enables extraction and moves disk I/O into the library, where a write failure fails the whole parse.
  - Perf red flags when enabled: duplicate detection is an O(n²) scan comparing raw multi-MB byte buffers, and the doc-global cache holds both raw and encoded bytes for every unique image until parse end, with several redundant full clones.
  - Schema drift: Node CLI always emits images: [] / imageErrorCount: 0 (camelCase) while the Rust CLI skips the keys entirely (snake_case) — for a SaaS source of truth, the two CLIs now produce differently-shaped JSON for the same document.
  - image_error_count increments from bounds-discovery failures even with extraction off, so a "new" key can appear in default output.

hexapode:
Superseded by #374, which consolidates all optional extraction outputs on current main with one computation-gated options flow and one CLI JSON schema.

PR #372, opened 2026-07-21
Add opt-in vector graphics extraction

hexapode:
## Summary

- add default-off `extract_vector_graphics` / `extractVectorGraphics` / `--extract-vector-graphics` across Rust, Rust CLI single+batch, NAPI/Node API+CLI, PyO3/Python API+CLI, and WASM
- expose page-scoped shapes and merged horizontal/vertical line segments without changing the existing internal lossy layout primitives
- preserve nested Form XObject transforms and model LlamaParse PDFium paint/color/curve and line-merging semantics
- omit vector output entirely by default to avoid payload growth
- document the API and the `bbox`/`width`/`height` naming differences from LlamaParse PDFium `coords`/`w`/`h`

## Validation

- `cargo test -p liteparse --no-default-features --lib --bins` — 258 library + 2 CLI tests passed
- `cargo test -p liteparse-pdfium nested_form_matrix_composition` — passed
- focused core extraction/merge/serde tests — 4 passed
- focused NAPI binding conversion test — passed
- focused PyO3 binding conversion test — passed
- `cargo check -p liteparse-napi -p liteparse-python --no-default-features` — passed
- `cargo check -p liteparse-wasm --target wasm32-unknown-unknown` — passed
- `pnpm --dir packages/node build:ts` — passed
- Python `compileall` — passed
- real-PDF CLI smoke with OCR disabled: default output omitted `vector_graphics`; enabled output returned the expected shape and merged line with bbox, ARGB color, and stroke width

The full integration run passed 9/11 tests. The two Office conversion tests fail because the locally installed LibreOffice process aborts in this sandbox; all PDF/image integration tests pass.

## Representation notes

Like the current LlamaParse PDFium extractor, public `lines` contain only horizontal and vertical segments. Diagonal and Bezier geometry remains represented by its parent shape (`bbox` plus `has_curve`) rather than raw path commands. LiteParse does not add PDFium extract-binary filesystem output because vector data is structured page metadata, not a binary artifact.


logan-markewich:
fyi, this change creates a second, parallel line-extraction semantics next to the internal layout primitives (H/V-only, ignores close-flags, own tolerances — the two can silently disagree about what lines exist on a page, and unifying later with table detection will be a project); output is unbounded per-page arrays (a CAD/matplotlib page can produce MBs); and the bindings introduce three rect representations (JsRect f64, VectorRect f32, and the pip wrapper flattens bbox to a positional tuple while PyO3 users get an object — that one I'd fix before merge).

hexapode:
Superseded by #374, which consolidates all optional extraction outputs on current main with one computation-gated options flow and one CLI JSON schema.

PR #381, opened 2026-07-27
feat: expose PDF document provenance metadata

hexapode:
## Summary
- expose document provenance metadata through Rust, Node.js, Python, and WASM parse results
- collect PDF dates, version and security details, signature state, incremental-save markers, trailer IDs, XMP, and source size
- preserve the default CLI JSON shape while supporting both path and byte inputs

## Validation
- cargo check for core, PDFium, NAPI, Python, and WASM without Tesseract
- focused document metadata unit tests, including scan chunk boundaries
- PDF path and byte-input integration tests
- cargo fmt and git diff checks
- Python source syntax validation


logan-markewich:
Noticing a few issues with this PR, going to push some changes shortly

crates/liteparse/src/document_metadata.rs
```diff
@@ -0,0 +1,257 @@
...
+const SCAN_BUFFER_BYTES: usize = 1 << 20;
+const SCAN_OVERLAP: usize = 64;
+const XMP_MAX_BYTES: usize = 64 * 1024;
+/// Above this size, resolving the catalog's `/Metadata` object costs more than
+/// the whole parse: lopdf decodes every stream, so a 103 MB file takes ~7.8 s
+/// (worst case measured under the cap is ~0.5 s). Larger documents report no
+/// `xmp` at all rather than a cheaper guess — the first `<?xpacket` in the raw
+/// bytes is frequently an embedded image's packet, not the document's.
+#[cfg(not(target_arch = "wasm32"))]
+const XMP_CATALOG_MAX_FILE_BYTES: u64 = 16 * 1024 * 1024;
```
logan-markewich, line 16: @hexapode note this limitation. I can remove it, but some files are very very slow to extract

logan-markewich (approved):
Made a few fixes. I think this looks good but its worth a second pass on your end @hexapode

PR #386, opened 2026-07-29
Fix parsing of centered borderless tables

ShiyuanHsu:
## Summary

- Preserve borderless table regions through XY-cut when rows share stable centered column tracks.
- Let the existing Markdown table classifier process the intact rows.
- Add regression tests for two-row tables and protection against vertical table slicing.

## Root cause

For centered cell text, varying text widths change the left edges while the column centers remain aligned. XY-cut interpreted the whitespace between those columns as page-column gutters and split one logical table into separate regions. The Markdown classifier then received column-wise text instead of complete table rows.

## Validation

- `cargo test -p liteparse --no-default-features --lib centered_text_table_obstacle -- --nocapture`
- Manually verified local parsing of native-text PDF samples without OCR.

### DIFF
--- a/crates/liteparse/src/projection.rs
+++ b/crates/liteparse/src/projection.rs
@@ -2881,9 +2881,18 @@ pub fn project_pages_to_grid(pages: Vec<Page>) -> Vec<ParsedPage> {
                 page.page_width,
                 page.page_height,
             );
-            let mut obstacles: Vec<Rect> = Vec::with_capacity(figures.len() + table_rects.len());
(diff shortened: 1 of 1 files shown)


logan-markewich:
For changes to markdown, I would recommend re-running [ParseBench](https://github.com/run-llama/ParseBench) -- we have tracked results there so its worth checking if this impacts performance there. I can always run it too if thats too annoying. I usually just point claude code or codex at it and let it wire up a local build for testing

PR #388, opened 2026-07-30
test(pdfium): add LI-8306 Form-/BBox white-render fixtures

SebasGarcia08:
## What

Test-only regression fixtures for the LI-8306 PDFium white-render bug: a Form XObject whose /BBox contains an out-of-range integer (e.g. 2^32) has the coordinate silently collapse to 0 in stock PDFium, producing an empty clip and a blank page.

## Fixtures (crates/pdfium/tests/li8306_form_bbox_render.rs)

- **2^32 out-of-range /BBox must render visibly** — `#[ignore]`d until a pdfium-binaries release carries the LI-8306 provenance fix (`PDFIUM_RELEASE_TAG` bump past chromium/7897); verified green against a patched candidate build on native ARM64 Linux.
- **Genuine zero /BBox stays an empty clip** (legacy behavior preserved).
- **Valid /BBox control renders visibly.**
- **Linux-only identity test** reports the mapped libpdfium.so path + blake3 from /proc/self/maps, asserting it when `LI8306_EXPECT_LIBPDFIUM_BLAKE3` is set.

## Scope

Test-only: no production code, lock, or renderer changes; the global PDFium lock is untouched. A candidate library is selected via the existing `PDFIUM_LIB_PATH`/`PDFIUM_INCLUDE_PATH` build-time override. CI-safe: the overflow assertion is ignored against the stock library and passes `--include-ignored` against the patched candidate (4/4 on native aarch64).

Related: run-llama/platform#22997 (Platform E2E), Linear LI-8306.

### DIFF
--- a/crates/pdfium/tests/li8306_form_bbox_render.rs
+++ b/crates/pdfium/tests/li8306_form_bbox_render.rs
@@ -1,0 +1,148 @@
+//! LI-8306 (test-only): Form XObject /BBox white-render regression fixtures.
+//!
+//! Renders a synthetic PDF whose Form XObject /BBox contains an out-of-range
+//! integer (2^32). Unpatched PDFium collapses the coordinate to 0, producing
+//! an empty clip and a blank page; the provenance-preserving build recovers
+//! the saturated coordinate and renders visibly. Zero and valid controls pin
+//! the unchanged legacy behavior.
+//!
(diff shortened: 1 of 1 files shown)


SebasGarcia08:
Closing for now — will re-propose later in public-appropriate form.

Issue #363, opened 2026-07-18
[Bug] get_config() silently omits 8 newer LiteParse constructor options


ASHUTOSH-KUMAR-RAO:
### Description

`liteparse.LiteParse.__init__()` accepts 8 newer constructor options (`image_mode`, `extract_links`, `ocr_failure_fatal`, `ocr_hedge_delays_ms`, `emit_word_boxes`, `crop_box`, `skip_diagonal_text`, `include_complexity`), but `get_config()` returns a `LiteParseConfig` object that omits all 8 fields. The settings are applied correctly internally, but invisible when inspecting the config — no error or warning is raised, they're just silently missing.

### Steps to Reproduce

1. Create a parser with newer options: `LiteParse(emit_word_boxes=True, skip_diagonal_text=True, include_complexity=True, extract_links=False)`
2. Call `parser.get_config()` and print the result
3. Check `hasattr(config, "emit_word_boxes")` etc. — all return `False`

Full repro script:

​```python
from liteparse import LiteParse

parser = LiteParse(
    emit_word_boxes=True,
    skip_diagonal_text=True,
    include_complexity=True,
    extract_links=False,
)

config = parser.get_config()
print(config)
print("emit_word_boxes in config?", hasattr(config, "emit_word_boxes"))
print("skip_diagonal_text in config?", hasattr(config, "skip_diagonal_text"))
print("include_complexity in config?", hasattr(config, "include_complexity"))
print("extract_links in config?", hasattr(config, "extract_links"))
​```

### Error Message

```text
No exception is raised — this is a silent omission, not a crash. Observed output:

LiteParseConfig(ocr_language='eng', ocr_enabled=True, ocr_server_url=None, ocr_server_headers=None, tessdata_path=None, max_pages=1000, target_pages=None, dpi=150.0, output_format='json', preserve_very_small_text=False, password=None, quiet=False, num_workers=11)

emit_word_boxes in config? False
skip_diagonal_text in config? False
include_complexity in config? False
extract_links in config? False

None of the 4 options passed to the constructor appear on the returned config object.
```

### LiteParse Version

2.6.0

### Operating System

Windows

### Node.js Version

N/A

### Additional Context

This is Python-specific — the Node.js bindings do not have this issue since they don't go through a similar config-conversion layer. Filed after digging through `packages/python/liteparse/parser.py` and `types.py` to compare the constructor's accepted kwargs against what `get_config()` actually returns.

[...]

PR #389, opened 2026-07-30
fix(python): expose render_form_fields in get_config()

shuvamk:
## Summary

`LiteParse.__init__()` accepts `render_form_fields` and forwards it to the
native parser, but the `LiteParseConfig` dataclass has no such field and
`get_config()` never copies it back. The option applies correctly — only the
reflected config denies it exists.

Same class as #363 (fixed in #368). Python-side only: the PyO3 getter already
exists on `PyLiteParseConfig`, so no `crates/` change is needed.

## Problem

```python
from liteparse import LiteParse

p = LiteParse(render_form_fields=True)

p._native.config.render_form_fields             # True  — the option did apply
hasattr(p.get_config(), "render_form_fields")   # False
p.get_config().render_form_fields
# AttributeError: 'LiteParseConfig' object has no attribute 'render_form_fields'
```

| | expected | actual on main |
|---|---|---|
| `hasattr(get_config(), "render_form_fields")` | `True` | `False` |
| `LiteParse(render_form_fields=True).get_config().render_form_fields` | `True` | `AttributeError` |
| `LiteParse().get_config().render_form_fields` | `False` | `AttributeError` |

No error or warning — the setting is silently invisible when inspecting the
config, exactly as described in #363.

## Cause

`render_form_fields` was added to the Python constructor and forwarded through
`kwargs` (parser.py), and `PyLiteParseConfig` exposes the getter
(`crates/liteparse-python/src/lib.rs`), but the wrapper's mirror of the option
set was not updated. On `main`, `PyLiteParseConfig` exposes 33 fields while
`LiteParseConfig` declares 32 and `get_config()` passes 32 — the symmetric
difference is exactly `{render_form_fields}`.

## Fix

Two lines: add `render_form_fields: bool = False` to `LiteParseConfig` and pass
`render_form_fields=cfg.render_form_fields` in `get_config()`. No behaviour
change to parsing, and no change to CLI or JSON output.

## Tests

Adds `packages/python/tests/test_config_e2e.py` covering the default (`False`)
and the constructor round-trip (`True`).

Verified both fail without the fix. With `types.py` and `parser.py` reverted to
`main` and only the test applied:

```
E   AttributeError: 'LiteParseConfig' object has no attribute 'render_form_fields'
tests/test_config_e2e.py:11: AttributeError
E   AttributeError: 'LiteParseConfig' object has no attribute 'render_form_fields'
tests/test_config_e2e.py:15: AttributeError
2 failed in 0.01s
```

Restored:

```
2 passed in 0.01s
```

Commands run:

- `python -m pytest tests/test_config_e2e.py` from `packages/python/` — 2 passed
- `SKIP_INTEGRATION_TESTS=yes cargo test --workspace --no-default-features --exclude liteparse-python --exclude liteparse-napi` — 306 passed, 0 failed (unchanged; no Rust touched)
- The `ci-python.yml` smoke shape (`from liteparse import LiteParse, ParseResult, ParsedPage, TextItem, ScreenshotResult`; construct; `repr`) — passes, and `get_config()` now ends `..., extract_content_bounds=False, render_form_fields=False)`

Two notes so nothing is overstated:

- I built the extension module locally with
  `cargo rustc -p liteparse-python --no-default-features -- -C link-arg=-undefined -C link-arg=dynamic_lookup`
  and dropped it in as `liteparse/_liteparse.so`, rather than building a wheel
  with maturin. A `pip install` of a real wheel was not exercised.
- The test has to be named explicitly. A bare `pytest tests/` aborts before
  reaching it: the three existing modules fail at collection on `main` already
  (`ImportError` on `BatchResult`, `BoundingBox`, `ImageFormat`, which
  `liteparse/__init__.py` no longer exports), giving
  `Interrupted: 3 errors during collection`. Their fixtures also point at
  `test-docs/` and `expected-dataset/`, which aren't in the repo.
  `ci-python.yml` never invokes pytest, so none of this is visible in CI. I've
  left it alone as a separate concern; the new module is self-contained and
  needs no fixture document.

### DIFF
--- a/packages/python/tests/test_config_e2e.py
+++ b/packages/python/tests/test_config_e2e.py
@@ -1,15 +1,66 @@
 """E2E tests for LiteParse.get_config() — validates the resolved config mirrors the native config."""
 
+import inspect
+from dataclasses import fields
+
 from liteparse import LiteParse
+from liteparse._liteparse import PyLiteParseConfig
+from liteparse.types import LiteParseConfig
+
+NATIVE_CONFIG_FIELDS = {
+    name for name in dir(PyLiteParseConfig) if not name.startswith("_")
+}
+
+NON_DEFAULT_OPTIONS = {
+    "ocr_enabled": True,
+    "ocr_server_url": "http://ocr.example/v1",
+    "ocr_server_headers": {"Authorization": "Bearer token"},
+    "ocr_language": "fra",
+    "tessdata_path": "tessdata",
+    "max_pages": 7,
+    "target_pages": "2-4",
+    "dpi": 222.0,
+    "output_format": "markdown",
+    "preserve_very_small_text": True,
+    "password": "s3cret",
+    "quiet": True,
+    "num_workers": 3,
+    "image_mode": "embed",
+    "extract_images": True,
+    "image_output_dir": "out/images",
+    "extract_links": False,
+    "keep_headers_footers": True,
+    "extract_annotations": True,
+    "extract_form_fields": True,
+    "extract_structure_tree": True,
+    "extract_xfa_packets": True,
+    "extract_content_bounds": True,
+    "detect_screenshot_rects": True,
+    "render_form_fields": True,
+    "ocr_failure_fatal": False,
+    "ocr_hedge_delays_ms": [0, 500],
+    "emit_word_boxes": True,
+    "extract_text_metadata": True,
+    "crop_box": (0.0, 0.25, 0.5, 0.125),
+    "skip_diagonal_text": True,
+    "include_complexity": True,
+    "extract_vector_graphics": True,
+}
 
 
-class TestGetConfig:
-    """Constructor options are reflected in the resolved config."""
+class TestConfigSurfaceParity:
+    """The Python config surfaces stay in step with the native config."""
 
-    def test_render_form_fields_defaults_to_false(self):
-        config = LiteParse(quiet=True).get_config()
-        assert config.render_form_fields is False
+    def test_every_python_config_surface_declares_the_native_fields(self):
+        assert {f.name for f in fields(LiteParseConfig)} == NATIVE_CONFIG_FIELDS
 
-    def test_render_form_fields_reflects_constructor_option(self):
-        config = LiteParse(render_form_fields=True, quiet=True).get_config()
-        assert config.render_form_fields is True
+        constructor_options = set(inspect.signature(LiteParse.__init__).parameters)
+        assert constructor_options - {"self"} == NATIVE_CONFIG_FIELDS
+
+        assert set(NON_DEFAULT_OPTIONS) == NATIVE_CONFIG_FIELDS
+
+    def test_get_config_reports_every_constructor_option(self):
+        config = LiteParse(**NON_DEFAULT_OPTIONS).get_config()
+
+        reported = {f.name: getattr(config, f.name) for f in fields(LiteParseConfig)}
+        assert reported == NON_DEFAULT_OPTIONS
--- a/packages/python/tests/test_config_e2e.py
+++ b/packages/python/tests/test_config_e2e.py
@@ -1,0 +1,15 @@
+"""E2E tests for LiteParse.get_config() — validates the resolved config mirrors the native config."""
+
+from liteparse import LiteParse
+
+
+class TestGetConfig:
+    """Constructor options are reflected in the resolved config."""
+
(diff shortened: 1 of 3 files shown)


packages/python/tests/test_config_e2e.py
logan-markewich, line 1: If we are going to test this, can we make it more thorough? Is there a test we can add that would have caught this ahead of time?

shuvamk:
Yes — a field-parity check over the whole config catches this class rather than just this one field. It needs two halves, because either one alone misses a case.

**1. Names** — `dir()` on the native config object vs `dataclasses.fields(LiteParseConfig)`. On `main` this reports exactly one gap and nothing else:

```
names missing from dataclass: ['render_form_fields']
```

**2. Values, with the option set to a non-default.** Half 1 only catches a field that never made it into `types.py`. If a field is declared in `types.py` but not passed through in `get_config()`, the dataclass default silently fills it in — I checked, and a default-constructed `LiteParse()` compares equal and passes even with the line deleted. Constructing with the option flipped and diffing each attribute does catch it:

```
value mismatches: [('render_form_fields', True, False)]
```

Both are green on this branch.

One thing worth knowing before deciding where to put it: **`packages/python/tests/` doesn't currently run anywhere.** `ci-python.yml` builds wheels and then runs an inline smoke/parse script — it never invokes `pytest`. And on `main`, `pytest tests/` aborts during collection:

```
ERROR tests/test_batch_e2e.py       - cannot import name 'BatchResult' from 'liteparse'
ERROR tests/test_parse_e2e.py       - cannot import name 'BoundingBox' from 'liteparse'
ERROR tests/test_screenshot_e2e.py  - cannot import name 'ImageFormat' from 'liteparse'
Interrupted: 3 errors during collection
```

(`conftest.py` also resolves fixtures from `test-docs/` and `expected-dataset/data`, neither of which is in the repo.) Because collection is interrupted, the file this PR adds isn't collected either under a bare `pytest tests/` — it only runs if you name it directly.

So the honest answer to "a test that would have caught this ahead of time" is: the parity check would have, but only if something runs it.

Tell me which you'd like and I'll push to this branch:

- replace the two field-specific tests with the parity check (keeps the PR small; still not gated by CI),
- that, plus a step in `ci-python.yml` that runs the file so it actually gates,
- or leave the three broken e2e modules to a separate PR — I didn't want to bundle unrelated repair into this one.

packages/python/tests/test_config_e2e.py
shuvamk: Agreed — replaced the two `render_form_fields` asserts with a parity test over the whole option set (`ae4f30b`).

`LiteParse.__init__`, `PyLiteParseConfig`, the `LiteParseConfig` dataclass and the `get_config()` body are four hand-mirrored declarations of the same options. The tests now assert:

1. all four declare the same 33 field names (native fields read off `dir(PyLiteParseConfig)`, so the native config stays the source of truth);
2. a parser built with a non-default value for every one of those options reports all 33 back through `get_config()` — that catches a field that is declared on the dataclass but never forwarded, which the name check alone would miss.

On `main` both fail, and the diff names exactly `render_form_fields`:

```
E       AssertionError: assert {'crop_box', ..._bounds', ...} == {'crop_box', ..._bounds', ...}
E         Extra items in the right set:
E         'render_form_fields'
```

On "would it have caught this ahead of time": I rebuilt the native module at `871697c` (the commit before #368) and ran the same test file there. It fails on the constructor check naming exactly the eight options #368 went on to add:

```
E       assert constructor_options - {"self"} == NATIVE_CONFIG_FIELDS
E       Extra items in the left set:
E       'extract_links'
E       'ocr_failure_fatal'
E       'image_mode'
E       'crop_box'
E       'skip_diagonal_text'...
```

So it covers both instances of this drift, not just the one field.

One caveat worth flagging: `ci-python.yml` doesn't invoke pytest, so this file doesn't gate anything in CI as things stand. Separately, `pytest packages/python/tests/` aborts on `main` with `Interrupted: 3 errors during collection` — `test_batch_e2e.py`, `test_parse_e2e.py` and `test_screenshot_e2e.py` fail to import `BatchResult`/`BoundingBox`/`ImageFormat` from `liteparse`, and their fixtures point at `test-docs/` and `expected-dataset/` which aren't in the repo. I ran the file by explicit path (`pytest tests/test_config_e2e.py`, 2 passed) against a locally built extension module. Happy to open a separate PR for the collection errors if that's useful, or to wire pytest into `ci-python.yml` — just didn't want to bundle either into this one.

PR #401, opened 2026-08-04
fix: include visible AcroForm values in text output

Neel49:
## Summary

Filled AcroForm values can be present in form fields but missing from LiteParse text and Markdown, making completed forms look blank.

This PR conditionally flattens visible form widgets before native text extraction, preserving their positions without using OCR. It also adds a regression test for filled, blank, and default-valued fields.


Neel49:
@codex review

PR #421, opened 2026-08-17
Preserve text in ambiguous table rows

Neel49:
## Summary
- Preserve all source text when a row has more cells than its inferred table columns.
- Use the existing GridFallback instead of dropping unmatched cells.
- Add regression coverage for nested label/value panels.

## Testing
- cargo fmt --all -- --check
- cargo clippy --all-targets
- cargo test -p liteparse


Neel49:
@codex review

logan-markewich:
Hey thanks @Neel49 -- can you try the latest 2.13.1 release? I made a similar fix in #420

PR #426, opened 2026-08-18
Expose configurable page screenshot output

adrianlyjak:
Adds packed RGB8 screenshot output and typed per-format screenshot options across Rust, Python, Node.js, and WASM.

Right now PNG output is a bottleneck for screenshot output speed if you don't want PNG output. This allows reading the direct image bytes with an external library (such as pillow) and have it output images directly in a specified format without the overhead of converting to PNG as a middleman.

End goal was jpg output. I experimented with using `image` to output the jpg directly, but unfortunately that's actually benched slower than handing off to pillow. 

API designed so that we can start defining different direct output image types, but kept scope minimal here. Added in a rudimentary png quality config in order to exercise the config API shape.


logan-markewich:
A few concerns flagged with this (apologies in advance for the claude-slop)

- `stride` is passed through from napi unnormalized, so PNG screenshots surface `stride: null` at runtime while the TypeScript type declares `stride?: number`
- Adding `#[serde(with = "serde_bytes")]` to `ScreenshotResult.image_bytes` silently changes the JS runtime type existing WASM consumers receive from `number[]` to `Uint8Array`, undocumented and inconsistent with the sibling `ExtractedImage.bytes`.
- With `detect_screenshot_rects` enabled, the rgb8 path materializes a full RGBA buffer and then a second full RGB buffer via `rgba_to_rgb`, so the 'raw pixels, no encode' format peaks at more memory than the PNG path it exists to beat.
- The publicly exported `ScreenshotResult` dataclass gained two required (default-less) fields, `format` and `stride`, inserted between `image_bytes` and the first defaulted field — a silent breaking change to a constructor signature.
- `getConfig()` returns `{ ...this._config }`, a shallow copy, so the newly added nested `screenshot` (and `screenshot.png`) object is shared by reference with the parser's internal cached config.
- `ScreenshotOptionsInput::extract` reads only the `format` and `png` keys and silently ignores every other key in the `screenshot` dict, so a typo'd option is a no-op instead of an error — unlike the other string options in this binding, which validate.
- `result.image_bytes.clone()` deep-copies the entire page raster for every screenshot; pre-existing, but 3-6x more expensive now that the payload can be raw RGB8 instead of a compressed PNG. Same pattern at crates/liteparse-wasm/src/lib.rs:1370.
- `screenshot.format` / `screenshot.stride` (and `r.format` / `r.stride` at lines 836-837, plus the three-level `cfg.screenshot.png.compression` chain at line 869) are read as bare attributes on the native object, while the very next line uses `getattr(screenshot, "is_solid_fill", False)` — this file uses defensive `getattr` for exactly this native/pure-Python skew.
 - `render_pages_to_png` keeps a body duplicated verbatim from the new `render_pages` (same `Library::init`, `load_document_from_input`, `render_document_pages`), differing only by passing `&ScreenshotOptions::default()` — and it does not pin `ScreenshotFormat::Png`, it inherits whatever the default happens to be.
- The new `screenshot` config was added to config.rs, parser.rs and all three binding crates, but no clap arg was added to `ScreenshotCommand`, so the option is unreachable from any CLI — skipping step 2 of the documented checklist.

Also, isn't stride fully computable from the width + format? It was a lot of churn to thread this through.

adrianlyjak:
> - stride is passed through from napi unnormalized, so PNG screenshots surface stride: null at runtime while the TypeScript type declares stride?: number
> - Adding #[serde(with = "serde_bytes")] to ScreenshotResult.image_bytes silently changes the JS runtime type existing WASM consumers receive from number[] to Uint8Array, undocumented and inconsistent with the sibling ExtractedImage.bytes.

These seem like good call outs

> With detect_screenshot_rects enabled, the rgb8 path materializes a full RGBA buffer and then a second full RGB buffer via rgba_to_rgb, so the 'raw pixels, no encode' format peaks at more memory than the PNG path it exists to beat.

The alternative is sharing a view which seems like a complex api to get right. While less than ideal, the bottleneck here the PNG (CPU) encoding

> The publicly exported ScreenshotResult dataclass gained two required (default-less) fields, format and stride, inserted between image_bytes and the first defaulted field — a silent breaking change to a constructor signature.

This is just a return type, no? Why would this be constructed by a client? Can make stride more optional

> - getConfig() returns { ...this._config }, a shallow copy, so the newly added nested screenshot (and screenshot.png) object is shared by reference with the parser's internal cached config.
> - ScreenshotOptionsInput::extract reads only the format and png keys and silently ignores every other key in the screenshot dict, so a typo'd option is a no-op instead of an error — unlike the other string options in this binding, which validate.

Will see if this can be improved. Idk if these are critical? Will look into it to understand better

> result.image_bytes.clone() deep-copies the entire page raster for every screenshot; pre-existing, but 3-6x more expensive now that the payload can be raw RGB8 instead of a compressed PNG. Same pattern at crates/liteparse-wasm/src/lib.rs:1370.

maybe can share memory better? For memory, this should ideally be done with bounded concurrency rather than calling to screenshot across a full doc at once to limit memory. I'm not sure about pushing concurrency semantics into the lib

> this file uses defensive getattr for exactly this native/pure-Python skew.

This seems weird? Why would we have skew? Is that something we intend to support? IMO that getattr crap should get ripped out

> render_pages_to_png keeps a body duplicated verbatim from the new render_pages (same Library::init, load_document_from_input, render_document_pages), differing only by passing &ScreenshotOptions::default() — and it does not pin ScreenshotFormat::Png, it inherits whatever the default happens to be.
> The new screenshot config was added to config.rs, parser.rs and all three binding crates, but no clap arg was added to ScreenshotCommand, so the option is unreachable from any CLI — skipping step 2 of the documented checklist.

👀 

> Also, isn't stride fully computable from the width + format? It was a lot of churn to thread this through.

Sort of agree here, but looked into it briefly and its a common return pattern in comparable libs, so seemed sort of nice? Can drop if you feel strongly https://pdfium.patagames.com/help/html/P_Patagames_Pdf_Net_PdfBitmap_Stride.htm

PR #428, opened 2026-08-19
Bound parse memory with configurable budgets and chunked OCR rendering

jonathan-whitaker:
## What changed

Two mechanisms so a single parse cannot grow without bound on pathological inputs:

- **`memory_budget_mb`** (default 4096, `0` disables): an approximate running bound on extracted content accumulated in memory — text items, word boxes, char codes, embedded image bytes, and screenshot PNGs share one pool. Crossing it fails the parse with the new typed `LiteParseError::MemoryBudget` (a clear, actionable message) instead of exhausting the process. Enforced in `parse()` and `is_complex()`. The estimate is a deliberate lower bound (payload bytes + struct sizes); vector graphics, structure-tree nodes, annotations, and form-field records are not counted, and the standalone `screenshot()` API is exempt (documented).
- **`ocr_raster_budget_mb`** (default 1024, `0` restores the previous behavior): OCR renders and recognizes pages in bounded rounds sized to this budget instead of pre-rendering every text-poor page's raster up front — peak raster memory is O(budget) instead of O(document). Measured on a 400-page sparse-text document at high render DPI: peak memory drops ~6x and the parse completes where the unbounded strategy exhausts a constrained container. Pages a prior extraction flattened are re-flattened per round on exactly those pages, preserving raster parity (including non-widget annotations on other pages). Rounds run serially, so effective OCR concurrency is `min(num_workers, pages per round)` — documented on the config field; the default admits far more pages per round than any realistic worker count.

Both knobs are exposed through the CLI (`parse` and `batch-parse`), Node (`memoryBudgetMb`/`ocrRasterBudgetMb`), Python, and WASM bindings. Serde defaults keep previously serialized configs working (absent keys get the defaults; explicit `0` is preserved).

## Breaking / semver

Requires a **minor** version bump: new `LiteParseError::MemoryBudget` variant, `LiteParseError` is now `#[non_exhaustive]`, and two new public fields on `LiteParseConfig`. In batch parsing the budget applies per batch (documented). A later-round OCR render failure now occurs after earlier rounds' recognitions have already run — relevant only to callers metering external OCR backends.

## Testing

- Unit + integration: budget trip fires per page with the running estimate reported; serde back-compat round-trips; round-boundary math (always admits one page; `0` = single round); the round test observes the round structure directly (distinct page sizes prove per-page OCR text routing; an in-flight high-water mark proves a one-page-round budget serializes recognition and `0` restores overlap); CLI flag parity test for both commands.
- Verified end-to-end in a downstream harness: bounded rounds hold a multi-hundred-page OCR-heavy parse ~6x lower in peak memory where the previous strategy was OOM-killed; the budget error surfaces cleanly across the napi boundary; output parity on regular documents is byte-identical.

### DIFF
--- a/crates/liteparse-napi/src/types.rs
+++ b/crates/liteparse-napi/src/types.rs
@@ -119,6 +119,14 @@ pub struct JsLiteParseConfig {
     pub include_complexity: Option<bool>,
     /// Expose page-scoped vector path extraction. Default false.
     pub extract_vector_graphics: Option<bool>,
+    /// Approximate memory budget (MiB) for extracted content accumulated
+    /// during a parse; the parse rejects with a clear error instead of
+    /// growing without bound. 0 disables. Default 4096.
+    pub memory_budget_mb: Option<u32>,
+    /// Approximate budget (MiB) for OCR page rasters held in memory at once;
+    /// pages render and recognize in rounds sized to this budget. 0 renders
+    /// all pages up front. Default 1024.
+    pub ocr_raster_budget_mb: Option<u32>,
 }
 
 /// A page sub-region as the fraction cropped from each side (top-left origin,
@@ -259,6 +267,12 @@ impl JsLiteParseConfig {
         if let Some(v) = self.extract_vector_graphics {
             cfg.extract_vector_graphics = v;
         }
+        if let Some(v) = self.memory_budget_mb {
+            cfg.memory_budget_mb = v as usize;
+        }
+        if let Some(v) = self.ocr_raster_budget_mb {
+            cfg.ocr_raster_budget_mb = v as usize;
+        }
         cfg
     }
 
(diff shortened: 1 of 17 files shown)


jonathan-whitaker:
Closing in favor of the streaming-first approach: callers bound result memory with the existing `parseBatches()` iterator, rather than adding new config options here.

For anyone pulling the bounded OCR rendering out of this branch separately, the pieces are independent of the `memory_budget_mb` option surface:

- `ocr_merge.rs`: `estimate_ocr_raster_bytes()` (round sizing that mirrors the long-edge DPI clamp) and the `flatten_page_numbers` parameter on `render_pages_for_ocr()` (per-page re-flatten so a freshly reopened document's rasters match what extraction saw, without hiding non-widget annotations on other pages).
- `extract.rs`: `flattened_page_numbers` on `ExtractedPages` (which pages extraction actually flattened — needed by the above).
- `parser.rs`: the round loop after the main critical section (`ocr_round_end()` + render/recognize/merge per round, document reopened per round — same tradeoff `parseBatches()` already makes per batch), replacing the render-all-pages-then-recognize pass.
- `tests/integration_test.rs`: `test_ocr_rounds_cover_every_page_once` — distinct page sizes prove per-page OCR routing across rounds, and an in-flight high-water mark proves a one-page-round budget serializes recognition while the single-round path overlaps.

The only coupling to config is the single `ocr_raster_budget_mb` value; it could just as well be a constant or derived from `num_workers`.

Issue #407, opened 2026-08-06
Proposal: opt-in structured layout blocks with bboxes and text-item provenance


catlan:
The markdown classifier's structure (regions, block types, table grids) is discarded at render time — callers get strings plus raw `textItems` with no link between them. That link can't be reconstructed downstream (whitespace collapse, de-hyphenation, wrapped joins, cell splitting); it only exists inside the classifier as blocks are formed. Exposing it enables grounded citations, layout-aware chunking, highlight overlays, and document comparison.

**Proposal:** a new flag in the `ExtractionOutputOptions` family (`emit_layout`? `extract_layout_blocks`? your call) adding `blocks` to each page:

```ts
type BBox = { x: number; y: number; width: number; height: number };

type LayoutBlock = {
  type: "paragraph" | "heading" | "list" | "table" | "pageFurniture";
  bbox: BBox;
  text: string;
  textItemIndices: number[];   // indices into the RETURNED page textItems
  level?: number;              // heading level / list nesting
  ordered?: boolean;           // list only
  marker?: string;             // list only
  furnitureKind?: "header" | "footer";
  table?: {
    cells: Array<Array<{
      text: string;
      bbox: BBox;
      textItemIndices: number[];
      isHeader: boolean;
    }>>;
  };
};
```

Contract:

- Off by default; all output byte-identical when off (verified as in #374).
- Indices refer to the returned page's `textItems` (post-projection, not paint order). One item may map to multiple cells when a run spanning columns is split.
- Table layouts are shape-validated at runtime; on mismatch the block demotes to `paragraph` (text + provenance intact) rather than shipping a malformed grid.
- `pageFurniture` = the lines the repetition filter removes; composes with `keep_headers_footers` (#382). Body blocks keep classifier reading order.
- Markdown rendering untouched — this is a parallel view of the same classification, so it inherits classifier improvements (#394 etc.) automatically.

We have this working on a fork, rebased on 2.11.1: aligned span→source-index vectors on `ProjectedLine`, provenance accumulated where the classifier forms/joins blocks, cell geometry threaded through all table paths including the new 2.11 ones. ~4k insertions in `crates/` incl. tests. Happy to PR it, whole or split (projection → classifier → tables → bindings).

Questions:

1. Naming/placement preference?
2. Which bindings in the first PR? WASM is complete; Node/Python DTOs are mechanical.
[...]

Issue #403, opened 2026-08-04
[Feature] Expose block-level structure with bounding boxes in JSON output


Keaanm:
### Problem Statement

The JSON output gives per-line textItems with bounding boxes, and the markdown
output gives block structure as text. Neither gives block structure with
coordinates.

If you need both, for example to map a chunk of text back to a region on the
page, you have to rebuild blocks yourself by clustering lines on vertical gaps.
That falls apart on pages with tight line spacing, where there is no gap to
split on. In those cases the markdown output shows plenty of headings, so the
classifier is clearly separating sections by font weight, size and indent, but
none of those signals are visible in textItems.

### Proposed Solution

Add a per-page `blocks` array to the JSON output, alongside `textItems`:

  {
    kind: "heading" | "paragraph" | "list_item" | "table" | "code" | "figure",
    level?: number,      // headings
    text: string,
    bbox: { x, y, width, height },
    page: number
  }

This is the decomposition `markdown_layout` already produces before rendering;
the request is to surface it rather than compute anything new.

### Alternatives Considered

I currently pull the heading lines out of the rendered markdown and force a
block break wherever one appears. That gets the section boundaries back, but it
only works for blocks that became headings. Paragraph breaks from bold
transitions, indent changes, list structure and column wrapping are still lost.
Matching heading text back to lines is also unreliable when words are
hyphenated or use ligatures.

Lowering the gap threshold does not fix it. On tightly spaced pages the line
spacing inside a paragraph is close to the spacing between them, so a threshold
low enough to split sections also shreds normal paragraphs.

### Additional Context

Tested on 2.11.0

logan-markewich:
2.13.0 added `extract-blocks` as an option, that should overall satisfy this
https://developers.llamaindex.ai/liteparse/guides/extraction/#layout-blocks

PR #430, opened 2026-08-21
feat: link extract_blocks output back to textItems

catlan:
Follow-up to #407 / #403. Adds `text_item_indices` to `LayoutBlock` and `LayoutCell` — indices into the returned page `textItems` — so callers can map blocks back to the items they already have (grounded citations, highlight overlays, layout-aware chunking, document comparison).

- Indices are recorded where blocks are formed: projection keeps a span→item alignment on `ProjectedLine`, and the classifier and table paths carry it through wraps, de-hyphenation, merges, and header folds. Cell indices are in reading order; block indices are the sorted union of their lines/cells.
- All bindings: core (and therefore CLI serde), wasm, napi, and python, including the public DTOs.
- `extract_blocks` now enables word-box extraction in core — the table detector needs real word geometry — and each binding serializes `words` only when the caller asked for them, so payloads are unchanged. (napi `parse_pages()` is untouched; the caller supplies geometry there.)
- With `extract_blocks` off, output is identical to v2.13.1; with it on, the new fields are additive.

### DIFF
--- a/crates/liteparse-napi/src/types.rs
+++ b/crates/liteparse-napi/src/types.rs
@@ -866,6 +866,9 @@ impl JsDocumentAnnotation {
 pub struct JsLayoutCell {
     pub text: String,
     pub bbox: Option<JsAnnotationRect>,
+    /// Indices into the page's returned `textItems`, in reading order, never
+    /// repeating within one cell; empty for padding cells.
+    pub text_item_indices: Vec<u32>,
 }
 
 /// A classified block plus where it sits on the page. Flat by design — `kind`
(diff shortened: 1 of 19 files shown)


logan-markewich:
We do we need to link back to textItems? The new blocks that got added already have the content + bounding boxes?

To me, these are two separate streams of usage/features right?

catlan:
Text + bbox get you display, but not mapping block content back onto the `textItems` the caller already has, and that mapping can't be rebuilt downstream:

- `text` is rendered output: whitespace collapsed, hyphenation dissolved, wrapped lines joined, straddling table runs split at estimated positions. String-matching it against `textItems` breaks exactly where the classifier did work.
- bbox containment is ambiguous in the hard cases: tight multi-column layouts, an item feeding two cells, padding cells with no ink. Recovering the assignment geometrically amounts to re-running a worse version of the classifier.

With the indices, a caller can anchor on the items: highlight a substring of a block (the block bbox is the whole paragraph; the items carry the fine geometry, plus word boxes when requested), item-granular citations, or chunking that keeps per-item metadata attached. Our concrete case is document comparison: word-level diff ops between two PDFs anchor to items through these indices; text+bbox alone couldn't place them.

The classifier already knows the mapping while it forms blocks, so this just stops discarding it. Cost is a vec of ints per block/cell, only when `extract_blocks` is on.

catlan:
@logan-markewich Anything I can answer / do, to get this merged?

PR #461, opened 2026-09-14
feat: use latest pdfium lib

logan-markewich:
Should be ~20% speedup in pdfium usage


logan-markewich:
paddle issue requires a separate fix

